PR #13, opened 2017-10-08
Default behavior

### LINKED ISSUE #2: Switch default behaviour
1. 404 as default behavior
2. www. redirect as default, when enabled via type www
3. redirect to $page as default, when enabled via type redirect with $page as option

alarsyo:
Fixes #2

### DIFF
--- a/caddy/caddy.go
+++ b/caddy/caddy.go
@@ -16,35 +16,43 @@ func init() {
 	})
 }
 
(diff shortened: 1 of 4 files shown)


stp-ip:
Looks good. Will deploy it this week and add all reposeed files for a v0.1.0 release too.
Then we can close the documentation issue and move to path based routing.

PR #18, opened 2017-10-11
Add Reposeed generated files

### LINKED ISSUE #6: Document relative vs absolute url usage for "to="
To=about.example.com will redirect to vhost/about.example.com.

Either we need to document to actually add the protocol or we should default the protocol to https://, when there is no forward slash at the beginning.

Maybe parse the "to" value and see, if it's an address or just a path.

alarsyo:


stp-ip:
Close to Github updating issues.

PR #31, opened 2018-01-09
Add basic Makefile to build caddy with txtdirect

### LINKED ISSUE #29: Move binary build logic into makefile
feature

**What happened**:
Currently binary build logic is only available within the travis script.

Separate the build logic into a makefile.

Might use: https://github.com/thockin/go-build-template

alarsyo:
Adds a Makefile to allow building locally easily
**Which issue this PR fixes**
Fixes #29 
This is not ready for merge.
**Release note**:

```release-note
Add Makefile to project
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 txtdirect
+caddy-copy/
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+txtdirect
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,4 @@
+FROM alpine:3.6
+RUN apk --no-cache add ca-certificates
+ADD txtdirect /caddy
+CMD ["/caddy"]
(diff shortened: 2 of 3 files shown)


alarsyo:
For now this is without the template suggested in #29, as the template would need modifications to integrate the caddy build logic. This more of a PoC for the moment.

stp-ip:
Going through the code that changes the local caddy code within the GOPATH. That might have problematic consequences. Might be necessary to vendor caddy code before that and then built from the vendored code. Thoughts?

alarsyo:
Right, didn't think about this! I'll think about how we can handle that.

stp-ip:
The local caddy copy would work. Wondering, if there is a sane way to reuse vendoring logic. We could be seen as a library, which shouldn't use vendoring...

alarsyo:
Yeah the caddy plugin system doesn't make it easy to integrate :/

stp-ip:
Would you like this to be self contained PR or do you want to reuse this within the travis build already?

alarsyo:
I do want to use it in the Travis build ! It'll probably involve some bash variable gymnastics to handle/detect cases correctly but hey once it's done it's for the best

stp-ip:
OK let's merge this already and do a follow on PR for the travis change.

PR #38, opened 2018-05-04
WIP: Implement path based routing

### LINKED ISSUE #11: Path based routing + gometa considerations
pkg.txtdirect.org/caddy -> meta redirect to github.com/txtdirect/txtdirect with path /caddy

_redirect.pkg.txtdirect.org IN TXT "v=txtv0;from=/caddy;to=about.txtdirect.org;type=path"
_redirect.caddy.pkg.txtdirect.org IN TXT "v=txtv0;to=https://github.com/txtdirect/txtdirect;type=gometa"

Clear up "from", "to" usage/naming.
Subpaths need to be added, which might lead to ugly urls: pkg.txtdirect.org/caddy/caddy -> Syntax for reusing the query?
_redirect.caddy.pkg.txtdirect.org IN TXT "v=txtv0;to=https://github.com/txtdirect/txtdirect{uri};type=gometa"?

erbesharat:
Implement path based routing. 

fixes #11 

This PR is not ready yet, It's open for disscussen so if I didn't understand something correctly or if something doesn't make sense we can fix it before going further.
**Release note**:

```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -1,0 +1,26 @@
+/*
+Copyright 2017 - The TXTdirect Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package txtdirect
+
+import (
+	"net/http"
+	"strings"
+)
+
+func redirectPath(w http.ResponseWriter, req *http.Request, r record, host, path string) error {
+	pathSlice := strings.Split(path, "/")
+	url := basezone + "." + pathSlice[1] + "." + host
+	http.Redirect(w, req, url, 302)
+	return nil
+}
--- a/path_test.go
+++ b/path_test.go
@@ -1,0 +1,32 @@
+/*
+Copyright 2017 - The TXTdirect Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package txtdirect
+
+import (
+	"net/http/httptest"
+	"testing"
+)
+
+func TestRedirectPath(t *testing.T) {
+	req := httptest.NewRequest("GET", "https://example.com/caddy", nil)
+	w := httptest.NewRecorder()
+	r := record{
+		Version: "txtv0",
+		To:      "https://example.com/caddy",
+		Type:    "path",
+	}
+	redirectPath(w, req, r, req.Host, req.URL.Path)
+	resp, _ := w.Result().Location()
+	t.Error(resp.String())
+}
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -34,6 +34,7 @@ type record struct {
 	Code    int
 	Type    string
 	Vcs     string
+	From    string
 }
 
 // Config contains the middleware's configuration
@@ -166,6 +167,14 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 	if rec.Type == "gometa" {
 		return gometa(w, rec, host, path)
 	}
+
+	if rec.Type == "path" {
+		return redirectPath(w, r, rec, host, path)
+	}
+
+	// if rec.Type == "path" {
+	// 	return RedirectPath(w, rec, host, path)
+	// }
 
 	return fmt.Errorf("record type %s unsupported", rec.Type)
 }
--- a/caddy/caddy_test.go
+++ b/caddy/caddy_test.go
@@ -139,6 +139,19 @@ func TestParse(t *testing.T) {
 				Enable:   []string{"host"},
 			},
 		},
+		{
+			`
+			txtdirect {
+				enable path
+				redirect https://example.com
+			}
+			`,
+			false,
+			txtdirect.Config{
+				Redirect: "https://example.com",
+				Enable:   []string{"path"},
+			},
+		},
 	}
 
 	for i, test := range tests {
--- a/txtdirect_test.go
+++ b/txtdirect_test.go
@@ -98,6 +98,17 @@ func TestParse(t *testing.T) {
 			record{},
 			fmt.Errorf("unhandled version 'txtv1'"),
 		},
+		{
+			"v=txtv0;to=https://example.com/caddy;type=path;code=302",
+			record{
+				Version: "txtv0",
+				To:      "https://example.com/caddy",
+				Type:    "path",
+				Code:    302,
+				Vcs:     "git",
+			},
+			nil,
+		},
 	}
 
 	for i, test := range tests {


alarsyo:
Taking a look as soon as possible !

erbesharat:
Thanks, It's kinda messy and not complete right now 😄 Just wanted to be sure that I've understood the concept correctly or not. And also I wanted to first implement it without wildcards concept to easily get more familiar with the project.

txtdirect.go
```diff
@@ -103,11 +105,25 @@ func getBaseTarget(rec record) (string, int) {
 
 func getRecord(host, path string) (record, error) {
 	zone := strings.Join([]string{basezone, host}, ".")
+	from := 0
+	if path != "" {
+		zone, from, _ = zoneFromPath(host, path)
```
alarsyo: If I understood the code, zone is the variable with the new sub domain (so for `example.com/test1/test2` zone will be `test2.test1.example.com`)

I don't find the variable name very explicit ; but maybe it refers to something I'm just not familiar with :/

I also don't have a better suggestion right now, just bringing it up

alarsyo: Okay apparently we already used basezone for `"_redirect"` so I see where that came from. 
Don't know why I used that name though, probably suggested by Michael.

txtdirect.go
```diff
@@ -103,11 +105,25 @@ func getBaseTarget(rec record) (string, int) {
...
 
+	for i := 1; i <= from; i++ {
+		zoneSlice := strings.Split(zone, ".")
+		zoneSlice[i] = "_"
+		zone = strings.Join(zoneSlice, ".")
+		s, err = net.LookupTXT(zone)
+		if err != nil && i == from {
+			return record{}, fmt.Errorf("could not get TXT record: %s", err)
+		}
+	}
```
alarsyo: Currently this loop doesn't stop if it does find a TXT record, is that always what we want to do ? 
Or should we stop to the first matching one for our path ?

txtdirect.go
stp-ip: Zone is used in DNS quite a lot. I was more confused by "from".

txtdirect.go
stp-ip: Yes the loop currently doesn't stop. This leads to the last record found being used or at least returned.
What probably should happen is if type=host -> don't consider path based code. For type=path first request already went through fine, then iterate through the wildcards until we get a result. The most specific aka in this case the earliest would be used.

txtdirect.go
```diff
@@ -167,5 +183,34 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 		return gometa(w, rec, host, path)
 	}
 
+	if rec.Type == "path" {
```
stp-ip: This code would probably never be reached. Only the first request is type=path. After it traverses the wildcards and didn't find a type=host it should redirect to a default instead. This would mean that the return from getRecords is either type=host or a failure.

txtdirect.go
```diff
@@ -167,5 +183,34 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
...
+
+func reverse(input []string) {
+	last := len(input) - 1
+	for i := 0; i < len(input)/2; i++ {
+		input[i], input[last-i] = input[last-i], input[i]
+	}
+}
+
+func zoneFromPath(host string, path string) (string, int, error) {
+	match, err := regexp.Compile("([a-zA-Z0-9])\\w+")
```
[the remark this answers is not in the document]
stp-ip: Not sure on the why either. This hardcodes the regex in the function for now. Later it should be done as configuration within the type=path record itself. 

On the regex itself. `([a-zA-Z0-9]+)` would match everything already, why the addition of `\\w+`?
The current regex also doesn't process single digits aka "0", "1" etc.

txtdirect.go
```diff
@@ -103,11 +105,25 @@ func getBaseTarget(rec record) (string, int) {
 
 func getRecord(host, path string) (record, error) {
 	zone := strings.Join([]string{basezone, host}, ".")
+	from := 0
+	if path != "" {
+		zone, from, _ = zoneFromPath(host, path)
+	}
 	s, err := net.LookupTXT(zone)
-	if err != nil {
+	if err != nil && from < 1 {
```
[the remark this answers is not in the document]
alarsyo: I also almost commented on that but removed my comment by understanding it this way:

If it fails, then it's only a problem if we don't have any other zones to try (if `from == 0`); 
Not triggering the if means either:
 - it didn't fail
 - it failed but `from > 0` so let's wait for the wildcard testing before saying it doesn't work

Not sure that's the clearest way to do it but I found that it made more sense when put this way

txtdirect.go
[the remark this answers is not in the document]
alarsyo: This issue with `from` plus the 3 return variables (which are fine, but we should ask if they're necessary before using 3 return values most of the time. Can't remember where I read that...) lead me to propose this change:

`zoneFromPath` could return a slice of the "subzones", and an error.
The current `from` variable could be derived from the slice length (so even if we don't find a better name, it's more obvious what it is when it is declared as `from := len(zones)`

Then we can still assemble the final hostname from the original `host` value + the `zones` slice

txtdirect.go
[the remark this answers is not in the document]
stp-ip: Good point. I think I can see the reasoning. Will think it through with a fresh mind again.

txtdirect.go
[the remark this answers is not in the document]
stp-ip: Depending on the changes, I feel the zoneFromPath might need to be refactored anyway. Moving len etc. out simplifies the return values, but also adds more code to the calling function. Definitely a balance we have to discuss.

alarsyo:
I added on to your work here: https://github.com/alarsyo/txtdirect/commits/path

The commit addresses some of the code-flow problems:
 - Now we stop on the first hostname that returns a record
 - We test the full hostname before diving into wildcards

Feel free to pull the changes into this PR if you like them

stp-ip:
> I have a question about a possible scenario:
> 
> Say a GET request comes for example.com. TXT record is looked up and it's of type path. As the path is empty, this should redirect to a default option directly, right ?

I would say yes it should use the default option either the one provided within the path record or within the txtdirect config.

One could argue that the path is not empty as it's `/`. So it could try:
_redirect._.example.com <-- default to a wildcard with an "empty" path

Definitely something we should discuss. What feels the most logical and natural.

txtdirect.go
erbesharat: Huh, it just named it "from" because I've read this on the issue:
pkg.txtdirect.org/caddy/v1 (from=/$1/$2)
Didn't know what else to name the variable, so I used from :smile:

stp-ip: So the idea behind the `from` was a simplified way to provide how a path is handled. Basically `from` what to `to`. `from` basically would be a simplified version of `re` for regex.

These names are up for discussion. Thought they made sense, but maybe not ^^.

path_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{
+			"example.com",
+			"/1/2",
+			"_redirect.2.1.example.com",
+			nil,
+		},
+		{
+			"example.com",
+			"/",
+			"_redirect.example.com",
```
stp-ip: "_redirect.example.com" would just lead to the path record. I think I suggested using "_redirect._.example.com", but that might not feel right.
Vague idea:
* if path == "/" + to != nil in `_redirect.example.com`-> rec.To
* if path == "/" + to == nil in `_redirect.example.com` -> rec.to from `redirect._.example.com

stp-ip: Does that make sense? @alarsyo

stp-ip: Mhh thinking about it. The general hierarchy of records with path in `example.com` would be:

specific zone www.example.com > wildcard zone _.example.com > default zone `to=` from .example.com

That would mean the above idea is incorrect as that would lead to the following hierarchy:
 default zone `to=` from .example.com > wildcard zone _.example.com

A few alternatives:
* use `default=` as the global path default in `example.com` <-- feels the most natural, would also be in line with the general host type I assume
* don't support an empty path redirect <-- probably the most common use case... so bad to not support it.
* do a workaround such as looking for _redirect._root.example.com <-- complex and unintuative

path.go
```diff
@@ -0,0 +1,53 @@
...
+	match, err := regexp.Compile("([a-zA-Z0-9]+)")
+	if err != nil {
+		return "", 0, err
+	}
+	pathSlice := match.FindAllString(path, -1)
+	from := len(pathSlice)
+	reverse(pathSlice)
+	url := append(pathSlice, host)
+	url = append([]string{basezone}, url...)
+	return strings.Join(url, "."), from, nil
```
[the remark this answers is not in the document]
erbesharat: I can't understand, can you please paraphrase

path_test.go
erbesharat: I made some changes in my last commit to fix this. Now if path if "/", it returns host's TXT record and then redirects to ``default=``.

path.go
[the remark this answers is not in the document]
stp-ip: Making the requested zone absolute (adding the trailing dot) should be done for all DNS requests not just in the preperation for path. (Not necessary in this PR, but in general we should move to always using absolute right away).

path.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"fmt"
+	"net"
+	"regexp"
+	"strings"
+)
+
+func zoneFromPath(host string, path string) (string, int, error) {
+	if path == "/" {
```
stp-ip: What's the reasoning behind this check here?

stp-ip:
Curious sidenote: /favicon.ico is currently also rewritten.

path.go
erbesharat: I did a double check and realized it's useless because at first we will get host's TXT record before calling this function, so there is no need to return host's zone in this function. Will remove it.

path.go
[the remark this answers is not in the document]
erbesharat: It happens for other requests in getRecord function, made this only for making path zones.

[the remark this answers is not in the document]
erbesharat: main.go - Line 103
It generates the zone file for host other types.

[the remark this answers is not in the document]
erbesharat: Ok, I just saw David's commit. will resolve the conflict.

erbesharat:
Should we add something like a blacklist slice in the code and redirect URLs like favicon.ico directly to the resource without checking the TXT record?

stp-ip:
A blacklist would probably be a good start. Not sure how better to solve this for now.

PR #46, opened 2018-05-29
Change DNS TXT lookup to use the absolute zone

### LINKED ISSUE #43: Use absolute zone with a trailing dot
feature

**What happened**:
Currently DNS requests don't use the absolute zone with a trailing dot. This might lead to longer request times as the whole dns search path might be tried depending on the setup.

Use _redirect.example.com. to always circumvent the searchpath.

Not sure, if that ever is a necessity to use the searchpath, but that can be reiterated later. For now it's unnecessary and actually makes requests much slower.

dbriesz:
Changes the DNS TXT lookup to add a trailing dot to the end of the zone in order to speed up lookup requests by avoiding the use of the entire DNS search path.

fixes #43

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -103,7 +103,13 @@ func getBaseTarget(rec record) (string, int) {
 
 func getRecord(host, path string) (record, error) {
 	zone := strings.Join([]string{basezone, host}, ".")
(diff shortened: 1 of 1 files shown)


stp-ip:
closing + reopen to rerun the flaky tests (network issue)

PR #47, opened 2018-05-31
Remove port if host includes one

### LINKED ISSUE #39: Zone shouldn't include a port, if host includes one
bug

**What happened**:
If the host variable has a port, it will be used in zone too and silently break.

Only use an actual zone, which doesn't include a port.

dbriesz:
Checks if the host variable contains a port and removes it before use in zone since adding a port to the zone will cause it to break.

fixes #39

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -14,6 +14,7 @@ limitations under the License.
 package txtdirect
 
 import (
+	"net/url"
 	"fmt"
 	"log"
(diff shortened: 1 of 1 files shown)


stp-ip:
Simple but in my eyes working solution. +1

stp-ip:
And we got the first merge from @dbriesz \o/

PR #53, opened 2018-06-13
Add tests for simplified regex

### LINKED ISSUE #50: [path] Add tests for simplified regex matching various path conventions
feature

**What happened**:
Currently we use a hardcoded simple regex.

Test many more path conventions to make sure we use the correct regex.

```
example.com/path/routing -> _redirect.routing.path.example.com
example.com/path-routing -> _redirect.path-routing.example.com
example.com/path-routing/test -> _redirect.test.path-routing.example.com
example.com/path_routing/test -> ?
example.com/special-chars/#?%! -> ?
example.com/path2routing/nested/test/ -> _redirect.test.nested.path2routing.example.com
example.com/123/test -> _redirect.test.123.example.com
example.com/path-routing?query=string -> _redirect.path-routing.example.com
example.com/path-routing?query=string&more=strings -> _redirect.path-routing.example.com
```

erbesharat:
Add tests for simplified regex

fixes #50 

**Release note**:
```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 )
@@ -24,12 +25,26 @@ func zoneFromPath(host string, path string, rec record) (string, int, error) {
 		if err != nil {
 			return "", 0, err
 		}
-		fromSlice := match.FindAllStringSubmatch(rec.From, -1)
-		generatedPath := []string{}
-		for _, v := range fromSlice {
+
+		fromSubmatch := match.FindAllStringSubmatch(rec.From, -1)
+		fromSlice := make(map[int]string)
+		for k, v := range fromSubmatch {
 			index, _ := strconv.Atoi(v[1])
-			generatedPath = append(generatedPath, pathSlice[index-1])
+			fromSlice[index] = pathSlice[k]
 		}
+
+		keys := []int{}
+		for k := range fromSlice {
+			keys = append(keys, k)
+		}
+		generatedPath := []string{}
+
+		sort.Sort(sort.Reverse(sort.IntSlice(keys)))
+
+		for _, k := range keys {
+			generatedPath = append(generatedPath, fromSlice[k])
+		}
+
 		url := append(generatedPath, host)
 		url = append([]string{basezone}, url...)
 		return strings.Join(url, "."), from, nil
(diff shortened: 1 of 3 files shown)


path.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func zoneFromPath(host string, path string, rec record) (string, int, error) {
-	match, err := regexp.Compile("([a-zA-Z0-9]+)")
+	match, err := regexp.Compile("[^\\/#\\.]+")
```
stp-ip: What's the specific reasoning behind this change? Was personally thinking about a different regex. `([^/]*)`

erbesharat: I've used this regex because browsers don't support # in URL if they are not at the end before a character. But I'm not sure about it. we can use the one you suggested if what I said isn't right.

stp-ip: So for a string like: `test/123#something`
Your regex would match 3 times: test, 123, something
The more general regex would just match 2 times: test, 123#something

Question would be, what would we expect?
In my view both seem slightly wrong. What we want would be `test`, `123` and then add the additional fragment to the resulting redirect, if enabled right? At least that would be my guess within a default one. Basically just use the path nothing more. Using fragments or query parameters could then be done via custom regex.

erbesharat: The only solution that I can think of right now is to split the path string into 2 parts from # and apply the regex only on the first part.

stp-ip: Going through this in my head. It seems we have to choices.

Only act on the path and therefore the regex only cares about the path.
Or we use a different regex. Let me check, if I can come up with one.

stp-ip: `\/([A-Za-z0-9-._~!$&'()*+,;=:@]+)` assuming I got the allowed chars right for paths.

We can then use grouping.
So the match could be /test, but the group we are actually interested in would be the following item in the given array.

Example: https://play.golang.org/p/slASFcOceO5

That way we get the match and the thing we actually want.

Does that make sense or am I missing something?

path_test.go
```diff
@@ -8,30 +8,85 @@ func Test_zoneFromPath(t *testing.T) {
...
 			"/",
+			"",
 			"_redirect.example.com",
 			nil,
 		},
+		{
+			"example.com",
+			"/caddy/pkg/v1",
+			"/$1/$3/$2",
+			"_redirect.caddy.v1.pkg.example.com",
```
stp-ip: Shouldn't this be: `_redirect.pkg.v1.caddy.example.com`

PR #54, opened 2018-06-28
Prevent internal folder path

### LINKED ISSUE #27: Prevent gometa with "internal" in path
feature

**What happened**:
Currently a folder path "internal" is not blacklisted.

Internal should be blacklisted and therefore prevent working around the possibility to import internal packages.

Set a vanity import to an internal package.

This is based on internal packages being introduced in 1.4 https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU

dbriesz:
Adds checks for paths that contain "internal" folders and returns an error since importing of internal packages should be prevented.

fixes #27

### DIFF
--- a/gometa.go
+++ b/gometa.go
@@ -14,6 +14,8 @@ limitations under the License.
 package txtdirect
 
 import (
+	"fmt"
+	"strings"
 	"html/template"
 	"net/http"
 )
@@ -28,6 +30,10 @@ var tmpl = template.Must(template.New("").Parse(`<!DOCTYPE html>
 func gometa(w http.ResponseWriter, r record, host, path string) error {
 	if path == "/" {
 		path = ""
+	}
+	bl := "/internal"
+	if strings.Contains(path, bl) {
+		return fmt.Errorf("path containing 'internal' is disallowed")
 	}
 
 	return tmpl.Execute(w, struct {
--- a/path.go
+++ b/path.go
@@ -8,10 +8,6 @@ import (
 )
 
 func zoneFromPath(host string, path string) (string, int, error) {
-	bl := "/internal"
-	if strings.Contains(path, bl) {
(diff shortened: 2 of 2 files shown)


gometa.go
```diff
@@ -29,6 +31,10 @@ func gometa(w http.ResponseWriter, r record, host, path string) error {
 	if path == "/" {
 		path = ""
 	}
+	bl := "/internal"
```
alarsyo: This is potentially me missing a really obvious thing, but I don't get what `bl` stands for :sweat_smile: depending on how obvious/non obvious it is we might want to change the naming ?

path.go
```diff
@@ -8,6 +8,10 @@ import (
 )
 
 func zoneFromPath(host string, path string) (string, int, error) {
+	bl := "/internal"
+	if strings.Contains(path, bl) {
+		return "", 0, fmt.Errorf("path containing 'internal' is disallowed")
+	}
```
alarsyo: Not sure if we want to check this here, as it would prevent a standard web redirect containing `/internal` in its path from being executed correctly. The place where you put it in `gometa` is enough, and catches it at the right moment, I would say.

gometa.go
stp-ip: blacklist.

path.go
dbriesz: Thanks - I removed that unnecessary check

stp-ip:
Looks good. Might need some tests.
@alarsyo @erbesharat thoughts?

alarsyo:
Yep, some tests in `gometa_test.go` and that should be good!

erbesharat:
LGTM
Just needs tests. :)

stp-ip:
Ok let's do tests in a follow-on PR @dbriesz

PR #55, opened 2018-06-29
Addresses multiple _redirect records and maximum _redirect length

### LINKED ISSUE #40: Only first TXT record is checked
bug

**What happened**:
TXTDirect fails if the DNS stores more than 1 record for a domain and the first one happens to not be the TXTDirect one.

TXTDirect should look for a record it recognizes among all the TXT records for a domain

In addition to the TXT record needed for TXTDirect, register other TXT records on the same domain

dbriesz:
Since we only check the first _redirect TXT record, if multiple _redirect records are provided, TXTDirect will only choose the first record. To address this issue, it is necessary to check the slice of TXT records and if it does not equal 1, we now return an error. 

Also, the maximum length of a _redirect record cannot exceed 255 characters. This PR addresses that issue as well by returning an error when a _redirect record exceeds 255 characters.

fixes #40

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -122,7 +122,7 @@ func getRecord(host, path string) (record, error) {
 	}
 
 	if len(s) != 1 {
-		return record{}, fmt.Errorf("could not parse TXT record with %d redirects", len(s))
+		return record{}, fmt.Errorf("could not parse TXT record with %d records", len(s))
 	}
 
 	rec := record{}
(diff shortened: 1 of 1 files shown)


stp-ip:
Feels too simple, but maybe that's correct.
@erbesharat @alarsyo thoughts?

alarsyo:
I'd also rebase the last commit as a fixup of the first one, and reword the first one now

erbesharat:
It should fix the problem, everything LGTM. Quick and clean fix. Good job David.

PR #56, opened 2018-07-04
Disable caddy telemetry

### LINKED ISSUE #48: Disable telemetry from caddy
feature

**What happened**:
Currently caddy introduced data collection.

Opt-in approach, with the default being metrics not being compiled in.

https://caddyserver.com/docs/telemetry#disable

dbriesz:
Disables caddy telemetry at compile time from TXTDirect so that data collection follows an opt-in approach.

fixes #48

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ build: fetch-dependencies
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/g" -- {} +
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/miekg\/caddy-prometheus"' -- {} +
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/captncraig\/caddy-realip"' -- {} +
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -10,6 +10,7 @@ build: fetch-dependencies
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/g" -- {} +
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/miekg\/caddy-prometheus"' -- {} +
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/captncraig\/caddy-realip"' -- {} +
+	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e 's/const enableTelemetry = true/const enableTelemetry = false/g' -- {} +
```
alarsyo: The `g` at the end of the regex isn't needed, it means `'global'`, so sed will execute the substitution for every match it finds in the file ; if you remove it it stops after the first one. 

Just a bit of sed/regex knowledge, in this case I don't think leaving the g or removing it would make a difference (and I see that some of the earlier sed commands have it as well even though it's not needed)

stp-ip:
Merging for now. Will most likely get refactored into the new makefile.

stp-ip:
Also sidenote. This missed the travis build. Aka the travis build would still have telemetry. Missed it myself before merge so. Follow-up PR please @dbriesz

PR #60, opened 2018-07-05
Disable caddy telemetry in Travis build

### LINKED ISSUE #48: Disable telemetry from caddy

dbriesz:
Disables caddy telemetry in Travis build. Also removes the unnecessary 'g' in the regex.

fixes #48

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,21 +6,22 @@ BUILD_REF=$(shell echo $(COMMIT) | cut -c1-6)
 build: fetch-dependencies
 	rm -rf caddy-copy
 	git clone https://github.com/mholt/caddy caddy-copy
-	find caddy-copy/caddyhttp/httpserver -name 'plugin.go' -type f -exec sed -i -e "s/gopkg/txtdirect/g" -- {} +
-	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/g" -- {} +
+	find caddy-copy/caddyhttp/httpserver -name 'plugin.go' -type f -exec sed -i -e "s/gopkg/txtdirect/" -- {} +
+	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/" -- {} +
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/miekg\/caddy-prometheus"' -- {} +
 	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/captncraig\/caddy-realip"' -- {} +
-	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e 's/const enableTelemetry = true/const enableTelemetry = false/g' -- {} +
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,6 +21,7 @@ travis-build: fetch-dependencies
 	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/g" -- {} + && \
 	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/miekg\/caddy-prometheus"' -- {} + && \
 	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/captncraig\/caddy-realip"' -- {} + && \
+	find caddy-copy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e 's/const enableTelemetry = true/const enableTelemetry = false/g' -- {} +
```
alarsyo:
```diff
-find caddy-copy/caddy/...
+find caddy/caddy/...
```

dbriesz: The build still fails - should line 24 of the makefile be changed to this?

```
find caddy/caddymain...
```

Makefile
```diff
@@ -6,21 +6,22 @@ BUILD_REF=$(shell echo $(COMMIT) | cut -c1-6)
...
 
 travis-build: fetch-dependencies
 	cd $$GOPATH/src/github.com/mholt/caddy && \
-	find caddyhttp/httpserver -name 'plugin.go' -type f -exec sed -i -e "s/gopkg/txtdirect/g" -- {} + && \
-	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/g" -- {} + && \
+	find caddyhttp/httpserver -name 'plugin.go' -type f -exec sed -i -e "s/gopkg/txtdirect/" -- {} + && \
+	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e "s/\/\/ This is where other plugins get plugged in (imported)/_ \"github.com\/txtdirect\/txtdirect\/caddy\"/" -- {} + && \
 	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/miekg\/caddy-prometheus"' -- {} + && \
 	find caddy/caddymain -name 'run.go' -type f -exec sed -i -e '/_ "github.com\/txtdirect\/txtdirect\/caddy"/a _ "github.com\/captncraig\/caddy-realip"' -- {} + && \
+	find caddy/caddy/caddymain -name 'run.go' -type f -exec sed -i -e 's/const enableTelemetry = true/const enableTelemetry = false/' -- {} +
```
alarsyo: For the Travis build we `cd` into the toplevel `caddy/` directory so we can remove one level of nesting from the paths; this one can be `caddy/caddymain` and the tests should pass

Makefile
alarsyo: Sorry, just saw your comment 😅

alarsyo:
Okay now we get another error in the build but I don't understand what's happening :/ has something changed on the Travis side ? Why would it fail to find the binary...

alarsyo:
Yeah this worked just fine on #64 this afternoon... Any clue ?

Makefile
dbriesz: No problem - glad we came to the same conclusion independently :smile:

dbriesz:
Maybe there's still a problem when removing the "g" from some or all of the regex lines?

alarsyo:
But in this case the build itself should fail, instead of erroring on the next command because it cannot find the output file 🤔

alarsyo:
Let's have @stp-ip take a look, maybe it's just too late and we're tired 😅

stp-ip:
Line 24. The added command doesn't chain to the next command aka ` && \` is missing.

PR #61, opened 2018-07-05
Add test for 'internal' folder in path

dbriesz:
Adds a test that ensures an error is returned when "/internal" is included in the path.

### DIFF
--- a/gometa_test.go
+++ b/gometa_test.go
@@ -84,3 +84,11 @@ func TestGometa(t *testing.T) {
 		}
 	}
 }
+
+func TestInternalFolderInPath(t *testing.T) {
+	rec := httptest.NewRecorder()
+	err := gometa(rec, record{}, "example.com", "/test/internal")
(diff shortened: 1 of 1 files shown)


gometa_test.go
```diff
@@ -84,3 +84,11 @@ func TestGometa(t *testing.T) {
 		}
 	}
 }
+
+func TestInternalFolderInPath(t *testing.T) {
+	rec := httptest.NewRecorder()
+	err := gometa(rec, record{}, "example.com", "test/internal")
```
alarsyo: The test would still work for catching whether internal is allowed or not, but for correctness `path` should be an absolute path, as it will be appended to a host (so `/test/internal` would be better)

alarsyo:
That commit can probably be rebased as a fixup as well, and then we can merge 😊

dbriesz:
Done :smile:

PR #64, opened 2018-07-06
Run tests during CI

### LINKED ISSUE #45: CI doesn't run tests
bug

**What happened**:

I just realized that Travis wasn't actually running any tests, just building caddy with TXTDirect inside.

Travis should run tests for TXTDirect in the "test" stage, as opposed to the "deploy" stage where building and uploading a Docker image is sufficient.

Trigger CI by opening a PR

alarsyo:

Previously we only tested if we could build correctly, now we also run tests.

**Which issue this PR fixes**
fixes #45

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ services:
   - docker
 
 script:
+  - go get -u github.com/mholt/caddy
+  - go test -v ./...
   - make travis-build
 
 stages:


alarsyo:
Woops, forgot that we needed to get the source before

.travis.yml
```diff
@@ -10,6 +10,7 @@ services:
   - docker
 
 script:
+  - go test -v ./...
```
stp-ip: This will most likely fail as the txtdirect plugin isn't yet linked. So it would need the linking as preparation step.

alarsyo: I don't think the linking to caddy is actually needed for our tests, they run fine on their own

alarsyo: At least they do on my machine

PR #65, opened 2018-07-09
Add documentation regarding encoding URLs

### LINKED ISSUE #62: Document URL format
feature

We need to document that URLs need to be escaped in the TXT record (especially the `;` character that would break parsing)

dbriesz:
Adds documentation regarding the need to encode URLs in the TXT record since not doing so can cause parsing issues and, as a result, break a user's configuration.

fixes #62

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -30,6 +30,20 @@ For configuration examples for the caddy plugin look at our [examples section](/
 ### Location
 The TXT record must be accessible under the subdomain "\_redirect".
 
+### URLs
+* Please encode all URLs in the TXT record since the parser expects encoded URLs. In some cases, unencoded characters may break your configuration.
+* Most notably, ";" should be escaped as "%3B" in a URL since the parser will split the URL on the unencoded semicolon.
+
+* Examples:  
+    ; -> %3B  
+    ? -> %3F  
+    = -> %3D  
+`https://example.com/already/encoded? -> https://example.com/already/encoded%3F`
+
+* External Links:  
+    https://en.wikipedia.org/wiki/Percent-encoding  
+    https://tools.ietf.org/html/rfc3986#page-11
+
 ### type=host
 *v*
 * Mandatory


docs/README.md
```diff
@@ -30,6 +30,18 @@ For configuration examples for the caddy plugin look at our [examples section](/
...
+* Please encode all URLs in the TXT record since the parser expects encoded URLs. In some cases, unencoded characters may break your configuration.
+* Most notably, ";" should be escaped as "%3B" in a URL since the parser will split the URL on the unencoded semicolon.
+
+Examples: <br />
+; -> %3B <br />
+/ -> %2F <br />
+? -> %3F <br />
+= -> %3D <br />
+
+*https://example.com/over/there?name=ferret -> https%3A%2F%2Fexample.com%2Fover%2Fthere%3Fname%3Dferret*
```
alarsyo: Maybe the users don't need to escape *everything* including the `https://` part and the `/` characters; we still want the records to be readable

docs/README.md
```diff
@@ -30,6 +30,18 @@ For configuration examples for the caddy plugin look at our [examples section](/
...
 
+### URLs
+* Please encode all URLs in the TXT record since the parser expects encoded URLs. In some cases, unencoded characters may break your configuration.
+* Most notably, ";" should be escaped as "%3B" in a URL since the parser will split the URL on the unencoded semicolon.
+
+Examples: <br />
+; -> %3B <br />
+/ -> %2F <br />
+? -> %3F <br />
+= -> %3D <br />
```
alarsyo: I may be wrong on that one but I think you can add two spaces at the end of a line to instruct markdown to add a linebreak. Allows the file to still be readable in text mode.

dbriesz: Thanks - that works

docs/README.md
dbriesz: Sounds good - I'm removing the URL scheme and `/` from what should be encoded. Should we add something to the docs that says that a `/` doesn't need to be encoded?

stp-ip:
After we agree on the final changes, we should probably rebase. Small change could be squashed/rebased to one commit or so.

On the change itself:
Might be good to link to more resources.
Either we can link to a more general overview such as https://en.wikipedia.org/wiki/Percent-encoding
Or the actual RFC section of relevance: https://tools.ietf.org/html/rfc3986#page-11

stp-ip:
Usually most software expects encoded urls so it's nothing new here just something that should be documented to prevent people running into issues.

dbriesz:
Sounds good and I'll do a rebase once the changes are finalized.

stp-ip:
Good for a rebase I think.

stp-ip:
OK wrong. Missed something.

docs/README.md
```diff
@@ -30,6 +30,20 @@ For configuration examples for the caddy plugin look at our [examples section](/
...
 
+### URLs
+* Please encode all URLs in the TXT record since the parser expects encoded URLs. In some cases, unencoded characters may break your configuration.
+* Most notably, ";" should be escaped as "%3B" in a URL since the parser will split the URL on the unencoded semicolon.
+
+* Examples:  
+    ; -> %3B  
+    ? -> %3F  
+    = -> %3D  
+`https://example.com/over/there?name=ferret -> https://example.com/over/there%3Fname%3Dferret`
```
stp-ip: This example url might be true or might not be.
Query strings are usually not encoded, if they are they are not query strings and are part of the url. This is a really special case and doesn't feel like the best example.

Better to use something like:
`https://example.com/already/encoded? -> https://example.com/already/encoded%3F`

dbriesz: Thanks for pointing that out - I'll change the example we use

PR #66, opened 2018-07-10
Fix getFinalRecord to find the final non-path record

### LINKED ISSUE #59: Chaining type=path results in error
bug

**What happened**:
Chaining multiple path routes together results in an error.

Is it necessary to be supported?

```
_redirect             60 IN TXT     "v=txtv0;to=http://root.worked.example.test;default=http://default.worked.example.test;type=path;code=302"

_redirect.two.one     60 IN TXT     "v=txtv0;to=http://two-one.worked.example.test;type=path;code=302"
```
example.test/one/two -> _redirect.two.one.example.test -> error

**Environment**:
- TXTDirect version: master

erbesharat:
Fix getFinalRecord to find the final non-path record

fixes #59 

**Release note**:

```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -79,7 +79,7 @@ func getFinalRecord(zone string, from int) (record, error) {
 		}
 		zone, from, err := zoneFromPath(url.Host, url.Path, rec)
 		if err != nil {
-			return rec, fmt.Errorf("could not generate zone URI from path. %s", err)
+			return rec, fmt.Errorf("could not generate zone URI from path: %s", err)
 		}
 		rec, err = getFinalRecord(zone, from)
 	}
--- a/path.go
+++ b/path.go
@@ -82,6 +82,9 @@ func getFinalRecord(zone string, from int) (record, error) {
 			return rec, fmt.Errorf("could not generate zone URI from path: %s", err)
 		}
 		rec, err = getFinalRecord(zone, from)
+		if err != nil {
+			return rec, fmt.Errorf("could not find the final record: %s", err)
+		}
 	}
 
 	return rec, nil
--- a/path.go
+++ b/path.go
@@ -3,7 +3,6 @@ package txtdirect
 import (
 	"fmt"
 	"net"
-	"net/url"
 	"regexp"
 	"sort"
 	"strconv"
@@ -73,18 +72,7 @@ func getFinalRecord(zone string, from int) (record, error) {
 	}
 
 	if rec.Type == "path" {
-		url, err := url.Parse(rec.To)
-		if err != nil {
-			return rec, fmt.Errorf("could not parse the url: %s", err)
-		}
-		zone, from, err := zoneFromPath(url.Host, url.Path, rec)
-		if err != nil {
-			return rec, fmt.Errorf("could not generate zone URI from path: %s", err)
-		}
-		rec, err = getFinalRecord(zone, from)
-		if err != nil {
-			return rec, fmt.Errorf("could not find the final record: %s", err)
-		}
+		return rec, fmt.Errorf("chaining path is not currently supported")
 	}
 
 	return rec, nil
--- a/path.go
+++ b/path.go
@@ -3,6 +3,7 @@ package txtdirect
 import (
 	"fmt"
 	"net"
+	"net/url"
 	"regexp"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


path.go
```diff
@@ -71,6 +72,18 @@ func getFinalRecord(zone string, from int) (record, error) {
...
 	}
 
+	if rec.Type == "path" {
+		url, err := url.Parse(rec.To)
+		if err != nil {
+			return rec, fmt.Errorf("could not parse the url: %s", err)
+		}
+		zone, from, err := zoneFromPath(url.Host, url.Path, rec)
+		if err != nil {
+			return rec, fmt.Errorf("could not generate zone URI from path. %s", err)
```
alarsyo: Small consistency issue with the error message following a period `.` instead of a colon `:` like in other error messages of the file

stp-ip:
Still run into issues with the code. Might be something on my side. Will do another debugging session after grabbing some :shallow_pan_of_food:

stp-ip:
Found a few other bugs while debugging this. Will need a fresh mind it seems. Will pick it up as soon as I can tomorrow.

stp-ip:
Current best guess:
Code is walking down the whole type=path based records, but doesn't stop at the correct time. It will grab or try to grab a host based one. If that's not possible the fallback code isn't triggered as it would in txtdirect.go.
That leads to an empty record and an error of `record type  unsupported`.

path.go
```diff
@@ -71,6 +72,18 @@ func getFinalRecord(zone string, from int) (record, error) {
...
+	if rec.Type == "path" {
+		url, err := url.Parse(rec.To)
+		if err != nil {
+			return rec, fmt.Errorf("could not parse the url: %s", err)
+		}
+		zone, from, err := zoneFromPath(url.Host, url.Path, rec)
+		if err != nil {
+			return rec, fmt.Errorf("could not generate zone URI from path: %s", err)
+		}
+		rec, err = getFinalRecord(zone, from)
```
erbesharat: I think it's because we don't return the error here on line 84. the fallback code only gets triggered when an error has been returned. am I right?

stp-ip:
Ok that partly fixes the issue.

The flow seems still wrong. But probably needs more discussion.

TXT record "example.test"
```
_redirect         60 IN TXT     "v=txtv0;to=http://fallback.worked.example.test;root=https://root.worked.example.test;type=path"

_redirect._       60 IN TXT     "v=txtv0;to=http://host.example.test;root=https://wildcard-root.worked.example.test;type=path;code=302"

_redirect.host    60 IN TXT     "v=txtv0;to=http://host.worked.example.test;type=host"
```

Visiting `example.test/test` leads to the following redirect flow:
* (backend) get `_redirect.example.test` - success
* (backend) get `_redirect.test.example.test` - no record
* (backend) get `_redirect._.example.test` - success
* (backend) get `_redirect.host.example.test` - success
* redirect to `fallback.worked.example.test`

From a flow perspective:
* why would it get `_redirect.host.example.test`?
* why does it redirect to `fallback.worked.example.test` aka rec.To from the first path record?

What do we want chaining to do exactly?
Going through your solution and the various test cases it might just not be an answered question yet.
Probably best to discuss this first and then see how we can fix it.
@alarsyo thoughts?

alarsyo:
I'm not sure I'm following what's the problem with the fact that it's getting `_redirect.host.example.test` :thinking: isn't it supposed to follow that, from the wildcard record ?

Also why does the wildcard need a `root` config as well ? wouldn't all requests matching the wildcard be redirected to host.example.test anyway ?

stp-ip:
> I'm not sure I'm following what's the problem with the fact that it's getting _redirect.host.example.test thinking isn't it supposed to follow that, from the wildcard record ?

Ok you are right it might make sense, but that probably depends on what we want chaining to do or if we just disallow it completely.

> Also why does the wildcard need a root config as well ? wouldn't all requests matching the wildcard be redirected to host.example.test anyway?

It doesn't need it. Maybe not the best idea to use a wildcard example, could have been just another `path` enabled redirect, which is unrelated and therefore has the root set.

Also in the mid term we want to support adding or piping data to the redirect such as `{url}` etc.

stp-ip:
I think we need to take a step back and think through the "problem" first. Will try to get that ordered in my head. Happy to take suggestions on chaining in general.

alarsyo:
I think chaining makes TXTDirect even more powerful, but it might make some complex configs harder to understand/debug in case something goes wrong. So definitely something think about; I don't know if your use cases would make heavy use of chaining or not

stp-ip:
Current use cases wouldn't use chaining. It came up as an edge case during testing. Definitely something that needs more use cases before we can say what makes sense. Considering this not high priority, but something to revisit and think about in parallel I think.

stp-ip:
So for this PR. How about we disallow chaining for now. Suggestion:
* get path record (as usual) -> used
* encounter another path record (chaining) -> not used
* log chaining currently not supported
* redirect to first path records fallback

That way we just fallback and have the ability to add chaining later on.

stp-ip:
Love how pushing something to be revisited simplifies a PR to a few lines.

@alarsyo another look please.

PR #73, opened 2018-07-12
Fix redirect loop when root key is not available in record

### LINKED ISSUE #70: Accessing the root page, without a "root" key configured ends in redirect loop
bug

**What happened**:
Accessing a root path via the path type ends in a redirect loop, if no "root=" is set on the record.

Fallback to the fallback path instead of ending in a redirect loop.

**Environment**:
- TXTDirect version: master

erbesharat:
Fix redirect loop when root key is not available in record

fixes #70 

**Release note**:

```release-note

```


stp-ip:
The latest addition looks good. Just wanted to wait for an additional review from @alarsyo on the various PRs. Shouldn't give a lot of merge conflicts so waiting shouldn't be an issue.

PR #74, opened 2018-07-14
Implement Custom DNS Resolver

### LINKED ISSUE #4: Ability to define specific dns resolver being used
In the long-term it might be interesting to specify an explicit dns resolver instead of using the system resolver. Might need a dns library change later on.

Not necessary for now, just wanted to add it to the backlog.

erbesharat:
Implement Custom DNS Resolver

fixes #4 

**Release note**:

```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -57,12 +57,12 @@ func getFinalRecord(zone string, from int, ctx context.Context, c Config) (recor
 	var txts []string
 	var err error
 
-	txts, err = net.LookupTXT(zone)
-
 	if c.Resolver != "" {
 		net := net.Resolver{}
 		net.Dial(ctx, "tcp", c.Resolver)
 		txts, err = net.LookupTXT(ctx, zone)
+	} else {
+		txts, err = net.LookupTXT(zone)
 	}
 
 	// if nothing found, jump into wildcards
--- a/path.go
+++ b/path.go
@@ -58,8 +58,7 @@ func getFinalRecord(zone string, from int, ctx context.Context, c Config) (recor
 	var err error
 
 	if c.Resolver != "" {
-		net := net.Resolver{}
-		net.Dial(ctx, "tcp", c.Resolver)
+		net := customResolver(c)
 		txts, err = net.LookupTXT(ctx, zone)
 	} else {
 		txts, err = net.LookupTXT(zone)
@@ -71,8 +70,7 @@ func getFinalRecord(zone string, from int, ctx context.Context, c Config) (recor
 		zoneSlice[i] = "_"
 		zone = strings.Join(zoneSlice, ".")
 		if c.Resolver != "" {
-			net := net.Resolver{}
-			net.Dial(ctx, "tcp", c.Resolver)
+			net := customResolver(c)
 			txts, err = net.LookupTXT(ctx, zone)
 		} else {
 			txts, err = net.LookupTXT(zone)
--- a/path.go
+++ b/path.go
@@ -70,11 +70,12 @@ func getFinalRecord(zone string, from int, ctx context.Context, c Config) (recor
 		zoneSlice := strings.Split(zone, ".")
 		zoneSlice[i] = "_"
 		zone = strings.Join(zoneSlice, ".")
-		txts, err = net.LookupTXT(zone)
 		if c.Resolver != "" {
 			net := net.Resolver{}
 			net.Dial(ctx, "tcp", c.Resolver)
 			txts, err = net.LookupTXT(ctx, zone)
+		} else {
+			txts, err = net.LookupTXT(zone)
 		}
 	}
 	if err != nil || len(txts) == 0 {
--- a/path.go
+++ b/path.go
@@ -1,6 +1,7 @@
 package txtdirect
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"regexp"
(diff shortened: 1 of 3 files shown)


path.go
```diff
@@ -49,18 +51,29 @@ func zoneFromPath(host string, path string, rec record) (string, int, error) {
 	return strings.Join(url, "."), from, nil
 }
 
-func getFinalRecord(zone string, from int) (record, error) {
+func getFinalRecord(zone string, from int, ctx context.Context, c Config) (record, error) {
 	var txts []string
 	var err error
 
 	txts, err = net.LookupTXT(zone)
```
alarsyo: Early feedback: the 'standard' lookup should only happen if the user didn't provide a custom resolver

erbesharat:
@stp-ip Can you check the code and test it when you have some time?

path.go
```diff
@@ -52,18 +54,29 @@ func zoneFromPath(host string, path string, rec record) (string, int, error) {
...
 	for i := 1; i <= from && len(txts) == 0; i++ {
 		zoneSlice := strings.Split(zone, ".")
 		zoneSlice[i] = "_"
 		zone = strings.Join(zoneSlice, ".")
 		txts, err = net.LookupTXT(zone)
+		if c.Resolver != "" {
+			net := net.Resolver{}
+			net.Dial(ctx, "tcp", c.Resolver)
+			txts, err = net.LookupTXT(ctx, zone)
+		}
```
stp-ip: Seems like this would do 2 different lookups. One for the standard resolver and one with the custom one. Or am I missing something?

txtdirect.go
```diff
@@ -217,8 +228,8 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 
 		if path != "" {
 			zone, from, err := zoneFromPath(host, path, rec)
-			rec, err = getFinalRecord(zone, from)
-			if err != nil {
+			rec, err = getFinalRecord(zone, from, r.Context(), c)
+      if err != nil {
```
stp-ip: Formatting seems broken here or is that just Github acting up.

path.go
```diff
@@ -74,6 +87,21 @@ func getFinalRecord(zone string, from int) (record, error) {
 		return rec, fmt.Errorf("could not parse record: %s", err)
 	}
 
+	if rec.Type == "path" {
```
stp-ip: Any specific reason to add path code in addition to the resolver changes?

erbesharat: I think I've messed up something when I rebased the branch.

txtdirect.go
erbesharat: I think formatting got broken when I used GitHub's editor to fix the conflict. Will fix it in the next commit.

path.go
erbesharat: We specify an alternate dialer on line 77 and use that instead of the default `net/http` package on line 78. Or at least that what I got from reading the `net/http`'s code.

stp-ip: Right, but in line 73 you are still using the old net.LookupTXT().

erbesharat: Ow, I missed that part and GitHub only shows line 76 to 79 :smile: my fault. will fix it right now.

stp-ip: No worries. That's what reviews are for. Just thought it might have been intentional.

stp-ip:
Code looks good. Will let merge once I gave it a test run.

alarsyo:
So after reading https://godoc.org/net#Resolver : the dial function you're calling is actually unimplemented by default, resulting in a nil pointer when you do call it :/ is seems we have no control over what actual DNS address is passed to `Dial`; what we can do is provide our own `Dial` which would ignore the address it's passed and always use the one we want instead. Example in this comment: https://github.com/golang/go/issues/19268#issuecomment-345328328

alarsyo:
Note that the proposal linked was denied, meaning the 'hack' of defining our own `Dial` that hardcodes the address of the DNS is the only way to do it for now

txtdirect.go
```diff
@@ -168,6 +178,17 @@ func fallback(w http.ResponseWriter, r *http.Request, fallback string, code int,
 	http.NotFound(w, r)
 }
 
+func customResolver(c Config) net.Resolver {
+	resolver := net.Resolver{
+		PreferGo: true,
+		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
+			d := net.Dialer{}
+			return d.DialContext(ctx, "tcp", c.Resolver)
```
alarsyo: Not sure we want to choose the networking type ourselves; DNS uses UDP by default if I recall correctly, and Go will probably choose for us when calling `Dial`

txtdirect.go
```diff
@@ -168,6 +178,17 @@ func fallback(w http.ResponseWriter, r *http.Request, fallback string, code int,
...
+		},
+	}
+	return resolver
```
alarsyo: Could be returned directly, don't know if this is optimized by the compiler.

stp-ip:
Preliminary tests worked.
CoreDNS using a text file for the records + TXTDirect with below Caddyfile.

```
example.test:80 {
  tls off
  txtdirect {
    enable path host
    resolver 127.0.0.1:53
  }
  log / stdout "{remote} - [{when_iso}] \"{method} {uri} {proto}\" {status} {size} {latency}"
  errors stdout
}
```

Awesome that should simplify local tests.

Agreeing with Antoine, that it needs a rebase. Will merge after WIP is removed from the title after rebase. \o/

Awesome team effort.

alarsyo:
That's not what I had in mind with a rebase :^)

PR #87, opened 2018-07-20
Add wrapper function for absolute zone and dialer code

### LINKED ISSUE #85: Absolute urls are not used on all dns requests
bug

**What happened**:
Currently not all dns requests use an absolute zone.

Use absolute zone on all dns requests

With the custom resolver and the absolute zone code, it might make sense to provide our own wrapper function for requesting dns records?

Related: https://github.com/txtdirect/txtdirect/blob/master/txtdirect.go#L128

**Environment**:
- TXTDirect version: master

dbriesz:
Adds a wrapper function so that all DNS requests use an absolute zone and also uses the custom or non-custom dialer where applicable.

fixes #85

### DIFF
--- a/path.go
+++ b/path.go
@@ -54,14 +54,9 @@ func zoneFromPath(host string, path string, rec record) (string, int, error) {
 }
 
 func getFinalRecord(zone string, from int, ctx context.Context, c Config) (record, error) {
-	var txts []string
-	var err error
-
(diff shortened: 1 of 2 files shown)


dbriesz:
:tada:

stp-ip:
I have to say with each PR the solutions get cleaner, easier and the commits more defined. Awesome.

PR #92, opened 2018-08-08
Implement logging

### LINKED ISSUE #37: Log all requests served
feature

**What happened:**
Especially with the failed implementation of the dockerv2 API, I think having a real log of each request processed and served would strongly benefit us for developping/debugging, and might be useful in production on the long term as well.

I guess caddy already logs incoming requests ? We probably need to record the redirects we send then, something like that. Thoughts ? Is it something that's already done for us completely in caddy, easily checked in real time from the built container ?

erbesharat:
Implements logging for redirects and incoming requests

fixes #37 

We need to talk about what we want to log and what informations we need to include in logs. Now, it only logs the incoming requests and redirects URL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ services:
 
 script:
   - go get -u github.com/mholt/caddy
+  - go get gopkg.in/natefinch/lumberjack.v2
   - go test -v ./...
   - make travis-build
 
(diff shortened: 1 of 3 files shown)


stp-ip:
I think each log line needs to be self contained or else you can't see which redirect was done by which request.
So something along the lines of `<timestamp> [txtdirect]: requestURL > 302 > redirectedURL`

stp-ip:
LGTM

stp-ip:
@dbriesz can take another look and then happy to merge.

PR #99, opened 2018-08-20
Refactor query to globally remove port from zone and use absolute zone

dbriesz:
Fixes a bug where port was being used with zone. Also uses query as a global function to remove port from zone and use absolute zone.

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -254,19 +254,16 @@ func query(zone string, ctx context.Context, c Config) ([]string, error) {
 		zone = zoneSlice[0]
 	}
 
-	var newZone string
 	if !strings.HasPrefix(zone, basezone) {
-		newZone = strings.Join([]string{basezone, zone}, ".")
-	} else {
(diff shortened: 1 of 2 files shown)


txtdirect.go
```diff
@@ -254,11 +248,25 @@ func customResolver(c Config) net.Resolver {
 }
 
 func query(zone string, ctx context.Context, c Config) ([]string, error) {
+	// Removes port from zone
+	if strings.Contains(zone, ":") {
+		zoneSlice := strings.Split(zone, ":")
+		zone = zoneSlice[0]
+	}
+
+	var newZone string
```
erbesharat: I think it's unnecessary to declare ``newZone``. Just check if ``zone`` has the ``basezone`` prefix, if yes put the result of ``strings.Join`` in ``zone``. Then there is no need to use ``else`` too.

stp-ip: +1

PR #100, opened 2018-09-04
Proxy requests instead of redirecting

### LINKED ISSUE #86: [proxy] proxy requests instead of redirecting
feature

**What happened**:
Currently we redirect the user, but don't enable proxying requests.

Ability to proxy to url instead of redirect.

Something to allow `demo.example.com` to be served from `example.com/demo` or `google.com` at `example.com`

* whitelist/blacklist urls (to prevent abuse).
* could be a new type or a specific config for host/path

Needs more discussion, but would be interesting.

dbriesz:
Adds a new proxy type and uses Caddy's reverse proxy functionality in order to perform proxy requests instead of redirecting.

PoC for #86

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -33,7 +33,9 @@ const (
 	basezone        = "_redirect"
 	defaultSub      = "www"
 	defaultProtocol = "https"
+	keepalive       = 30
 	logFormat       = "02/Jan/2006:15:04:05 -0700"
+	timeout         = 30 * time.Second
 )
 
 var PlaceholderRegex = regexp.MustCompile("{[~>?]?\\w+}")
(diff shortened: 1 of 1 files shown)


stp-ip:
Additional notes:

If we ever go with whitelist and more elaborate config, I think it needs to come from txtdirect config and not from DNS records. I'll post an example in the Proxy issue.

dbriesz:
Thanks - I added constants for keepalive and timeout (and set timeout to 30 * time.Second). It looks like the build passes without changing the makefile, but I'm happy to take another look :)

txtdirect.go
```diff
@@ -33,7 +33,9 @@ const (
 	basezone        = "_redirect"
 	defaultSub      = "www"
 	defaultProtocol = "https"
+	keepalive       = 30
 	logFormat       = "02/Jan/2006:15:04:05 -0700"
+	timeout         = 30 * time.Second
```
stp-ip: let's call them proxyTimeout and proxyKeepalive or else it's harder to distinguish.

dbriesz: Sounds good 👍

stp-ip:
Merging as unstable and undocumented PoC. Easier to iterate and test that way. Happy for a post-merge review @erbesharat @alarsyo

PR #102, opened 2018-09-11
PoC version of docker vanity urls

### LINKED ISSUE #14: Docker based container vanity urls
Example config for nginx: https://gist.github.com/thockin/b9fa587a63f4c642588813ab172e081d

Idea is to have a type: dockerv2. Redirecting a dockerv2 url would implement all redirects such as blobs, catalog etc.
First iteration would be similar to `type=host`. Basically providing simple non path based redirects for container images.

Optimally we would have `docker pull docker.txtdirect.org` working, but the registry needs a path image to work. So the client side call will look more like: `docker pull docker.txtdirect.org/txtdirect`

So the simplest case is probably force path to be a single directory first aka /txtdirect and have rec.to link to the namespace.

So `docker pull docker.txtdirect.org/txtdirect` with rec.to being `docker.io/seetheprogress`.

A few issues:
  * non empty path might conflict with the path type later on
  * workaround authentication with /v2/ being set to 200
  * authentication should be supported in a follow on PR
  * how would this work with path
  * how would this work with a gitlab registry, which can have subnamespaces

Gitlab registry idea:
`docker pull docker.txtdirect.org/namespace/subnamespace/imagename`
rec.To: `https://registry.gitlab.com/`
Result: `https://registry.gitlab.com/namespace/subnamespace/imagename`

erbesharat:
It defines a new type called ``dockerv2`` that handles docker vanity URLs redirection.

fixes #14 

It's jsut a PoC version and needs more work and discussion.

### DIFF
--- a/dockerv2.go
+++ b/dockerv2.go
@@ -36,7 +36,7 @@ func createDockerv2URI(to string, path string) (string, error) {
 		return "", err
 	}
 
-	if uri.Path == "/" {
+	if uri.Path == "/" || uri.Path == "" {
 		uri.Path = path
 		return uri.String(), nil
 	}
--- a/dockerv2.go
+++ b/dockerv2.go
@@ -16,9 +16,6 @@ var regexs = []string{
 
 func generateDockerv2URI(path string, rec record) (string, error) {
 	uri := rec.To
-	if rec.Re == "" && rec.To == "" {
-		log.Printf("<%s> [txtdirect]: re= and to= fields are required in dockerv2 type.", time.Now().Format(logFormat))
-	}
 	Dockerv2Regex, err := regexp.Compile(rec.Re)
 	if err != nil {
 		log.Printf("<%s> [txtdirect]: the given regex doesn't work as expected: %s", time.Now().Format(logFormat), rec.Re)
--- a/dockerv2.go
+++ b/dockerv2.go
@@ -2,11 +2,13 @@ package txtdirect
 
 import (
 	"fmt"
+	"net/http"
 	"regexp"
 	"strings"
 )
 
 var dockerRegexs = map[string]string{
+	"v2":        "/v2/?",
 	"_catalog":  "^/v2/_catalog$",
 	"tags":      "^/v2/(.*)/tags/(.*)",
 	"manifests": "^/v2/(.*)/manifests/(.*)",
@@ -15,7 +17,16 @@ var dockerRegexs = map[string]string{
 
 var DockerRegex = regexp.MustCompile("^\\/v2\\/(.*\\/(tags|manifests|blobs)\\/.*|_catalog$)")
 
-func generateDockerv2URI(path string, rec record) (string, int) {
+func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error {
+	path := r.URL.Path
+	v2Regex, err := regexp.Compile(dockerRegexs["v2"])
+	if err != nil {
+		panic(err)
+	}
+	if v2Regex.Match([]byte(path)) {
+		_, err := w.Write([]byte(http.StatusText(http.StatusOK)))
+		return err
+	}
 	if path != "/" {
 		regexType := DockerRegex.FindAllStringSubmatch(path, -1)[0]
 		pathRegex, err := regexp.Compile(dockerRegexs[regexType[len(regexType)-1]])
@@ -29,7 +40,9 @@ func generateDockerv2URI(path string, rec record) (string, int) {
 		for i, v := range pathSlice {
 			uri = strings.Replace(uri, fmt.Sprintf("$%d", i+1), v, -1)
 		}
-		return uri, rec.Code
+		http.Redirect(w, r, uri, http.StatusMovedPermanently)
+		return nil
 	}
-	return rec.To, rec.Code
+	http.Redirect(w, r, rec.To, http.StatusMovedPermanently)
+	return nil
 }
--- a/dockerv2.go
+++ b/dockerv2.go
@@ -1,0 +1,40 @@
+package txtdirect
+
+import (
+	"fmt"
+	"log"
+	"regexp"
+	"strings"
+	"time"
+)
+
+var regexs = []string{
+	"^/v2/_catalog$",
+	"^/v2/(.*)/tags/(.*)",
+	"^/v2/(.*)/manifests/(.*)",
+}
+
+func generateDockerv2URI(path string, rec record) (string, error) {
+	uri := rec.To
+	if rec.Re == "" && rec.To == "" {
+		log.Printf("<%s> [txtdirect]: re= and to= fields are required in dockerv2 type.", time.Now().Format(logFormat))
+	}
+	Dockerv2Regex, err := regexp.Compile(rec.Re)
+	if err != nil {
+		log.Printf("<%s> [txtdirect]: the given regex doesn't work as expected: %s", time.Now().Format(logFormat), rec.Re)
+	}
+	pathSubmatches := Dockerv2Regex.FindAllStringSubmatch(path, -1)
+	if len(pathSubmatches) < 1 {
+		log.Printf("<%s> [txtdirect]: can't extract submatches from %s, with %s", time.Now().Format(logFormat), path, rec.Re)
+	}
+	pathSlice := pathSubmatches[0][1:]
+
+	for i, v := range pathSlice {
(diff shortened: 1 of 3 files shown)


erbesharat:
@stp-ip So you mean we should have a collection of pre-defined regexes in our code that handle ``rec.to`` field for ``dockerv2`` type? That way we only need to remove the ``rec.re`` part and use the regexes that we've defined.

stp-ip:
Mhh not sure. Will clarify my thoughts once I can articulate them ^^

dockerv2.go
```diff
@@ -15,7 +17,16 @@ var dockerRegexs = map[string]string{
...
 var DockerRegex = regexp.MustCompile("^\\/v2\\/(.*\\/(tags|manifests|blobs)\\/.*|_catalog$)")
 
-func generateDockerv2URI(path string, rec record) (string, int) {
+func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error {
+	path := r.URL.Path
+	v2Regex, err := regexp.Compile(dockerRegexs["v2"])
+	if err != nil {
+		panic(err)
+	}
+	if v2Regex.Match([]byte(path)) {
```
stp-ip: The v2Regex matches all docker specific regexes and therefore returns StatusOK for all requests. Breaking the other redirects.

dockerv2.go
```diff
@@ -0,0 +1,49 @@
...
+	"tags":      "^/v2/(.*)/tags/(.*)",
+	"manifests": "^/v2/(.*)/manifests/(.*)",
+	"blobs":     "^/v2/(.*)/blobs/(.*)",
+}
+
+var DockerRegex = regexp.MustCompile("^\\/v2\\/(.*\\/(tags|manifests|blobs)\\/.*|_catalog$)")
+
+func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error {
+	path := r.URL.Path
+	v2Regex, err := regexp.Compile(dockerRegexs["v2"])
```
alarsyo: This could be precompiled as well, instead of compiling it everytime we call the function ?

I haven't tested this, but could it be possible to have something like this for `dockerRegexs` ?

```go
var dockerRegexs = map[string]Regexp{
	"v2":        regex.MustCompile("^\\/?v2\\/?$"),
	"_catalog":  regex.MustCompile("^/v2/_catalog$"),
	"tags":      regex.MustCompile("^/v2/(.*)/tags/(.*)"),
	"manifests": regex.MustCompile("^/v2/(.*)/manifests/(.*)"),
	"blobs":     regex.MustCompile("^/v2/(.*)/blobs/(.*)"),
}
```

The map *should* be able to contain any type we throw at it, I'm just not sure we can initialize it this way. (if we can't, maybe in an `init()` function ?

dockerv2.go
```diff
@@ -0,0 +1,49 @@
...
+		panic(err)
+	}
+	if v2Regex.MatchString(path) {
+		_, err := w.Write([]byte(http.StatusText(http.StatusOK)))
+		return err
+	}
+	if path != "/" {
+		fmt.Println(path)
+		regexType := DockerRegex.FindAllStringSubmatch(path, -1)[0]
+		pathRegex, err := regexp.Compile(dockerRegexs[regexType[len(regexType)-1]])
```
alarsyo: Maybe it's best to check if `regexType[len(regexType)-1]` has a value associated with it in the map, before passing the result to `regexp.Compile` ?

txtdirect.go
```diff
@@ -362,6 +365,14 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 		return nil
 	}
 
+	if rec.Type == "dockerv2" {
+		err := redirectDockerv2(w, r, rec)
+		if err != nil {
+			panic(err)
```
lbb: Please try to avoid panic. It seems that this will crash the program. Using panics and recover in an Exception style pattern is very expensive.

erbesharat: It's just for testing, will remove all the unnecessary logs and panics at the end.

dockerv2.go
```diff
@@ -0,0 +1,49 @@
...
+var DockerRegex = regexp.MustCompile("^\\/v2\\/(.*\\/(tags|manifests|blobs)\\/.*|_catalog$)")
+
+func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error {
+	path := r.URL.Path
+	v2Regex, err := regexp.Compile(dockerRegexs["v2"])
+	if err != nil {
+		panic(err)
+	}
+	if v2Regex.MatchString(path) {
+		_, err := w.Write([]byte(http.StatusText(http.StatusOK)))
```
stp-ip: Header is missing.
```
		w.Header().Set("Docker-Distribution-Api-Version", "registry/2.0")
```

stp-ip: *above the status being written

stp-ip:
So the above code changes makes it workable locally, but not yet with docker.io. It redirects weirdly.

Most of it can already be tested and debugged locally. Run a local docker registry https://docs.docker.com/registry/deploying/#run-a-local-registry
Redirect to that registry.
That way we can already work on getting the redirects and specific container redirects working.

Will jump on the plane now and not sure how far I can debug the docker.io issue there. Will keep this updated.

dockerv2.go
```diff
@@ -0,0 +1,43 @@
...
+
+func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error {
+	path := r.URL.Path
+	if dockerRegexes["v2"].MatchString(path) {
+		_, err := w.Write([]byte(http.StatusText(http.StatusOK)))
+		return err
+	}
+	if path != "/" {
+		regexType := DockerRegex.FindAllStringSubmatch(path, -1)[0]
+		if regexType[len(regexType)-1] == "" {
```
erbesharat: I've added this check and replacement code because of a weird bug in ``DockerRegex``.
In ``_catalog`` case, it returns an empty second group but in other languages like javascript, there isn't any second group. (Actually, there shouldn't be a second group at all). No idea why it happens in go :smile:

stp-ip: _catalog isn't used in the normal cases. It's hard to pinpoint, what's legacy and unused etc.
For now my tests have shown that _catalog and tags aren't really used, but let's keep them for now.

stp-ip:
Will be creating and deploying gcr.io for txtdirect this week so we can test with that and also use our own vanity url in production for txtdirect \o/ and txtd.io

stp-ip:
Still running into issues:
```
docker.example.test.            IN A 127.0.0.1
_redirect.docker.example.test.  IN TXT "v=txtv0;to=https://gcr.io;type=dockerv2"
```
There is still a race condition somewhere. I'm getting the following errors 70+% of the time instead of a successful pull:

docker log:
```
docker pull docker.example.test/google_containers/git-sync:v2.0.0
Error response from daemon: error parsing HTTP 404 response body: invalid character 'p' after top-level value: "404 page not found\n"
```

txtdirect log:
```
127.0.0.1 - [2018-11-25T20:41:37Z] "GET /v2/ HTTP/1.1" 200 2 415µs
127.0.0.1 - [2018-11-25T20:41:37Z] "GET /v2/google_containers/git-sync/manifests/v2.0.0 HTTP/1.1" 404 19 431µs
127.0.0.1 - [2018-11-25T20:41:38Z] "GET /v2/ HTTP/1.1" 200 2 435µs
127.0.0.1 - [2018-11-25T20:41:38Z] "GET /v2/google_containers/git-sync/manifests/v2.0.0 HTTP/1.1" 404 19 284µs
```

stp-ip:
This also needs tests. The various cases should be tested:
* full redirect
* specific redirect without version
* specific redirect with version
* alias redirect without version 
* alias redirect with version

dockerv2.go
stp-ip: Header is still missing.

dockerv2_test.go
```diff
@@ -0,0 +1,74 @@
...
+func Test_generateDockerv2URI(t *testing.T) {
+	tests := []struct {
+		path     string
+		rec      record
+		expected string
+	}{
+		{
+			"/v2/",
+			record{
+				To:   "docker.io/seetheprogress/txtdirect:latest",
```
stp-ip: let's test gcr.io only for now as docker.io doesn't work yet.

dockerv2.go
erbesharat: Probably got removed by hitting ctrl+z too many times :laughing:

PR #103, opened 2018-09-11
Refactor placeholders and support placeholders in dns records

### LINKED ISSUE #91: [placeholder] Placeholder specific redirects
feature

**What happened**:
Currently placeholders can be used to add information to the redirected url.

Use placeholder information to handle the redirect especially.
Such as `from={method}` could lookup:
```
_redirect.get.example.com
_redirect.post.example.com
```

Needs discussion. Just a thought and most likely not that useful, just wanted to get the idea out.

erbesharat:
Puts placeholder code in a separate file and uses ``parsePlaceholders`` in record parse method to support placeholders inside TXT records too.

fixes #91 

**Release note**:

```release-note

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ services:
 
 script:
   - go get -u github.com/mholt/caddy
-  - go get gopkg.in/natefinch/lumberjack.v2
+  - go get -u github.com/natefinch/lumberjack
   - go test -v ./...
   - make travis-build
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ services:
 script:
   - go get -u github.com/mholt/caddy
   - go get -u github.com/natefinch/lumberjack
-  - go test -v ./...
+  - go get -u github.com/miekg/dns
+  - go test -v -race ./...
   - make travis-build
 
 stages:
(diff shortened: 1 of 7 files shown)


erbesharat:
@stp-ip I just think it would be a good idea to let the user use placeholders in records too. Have no idea if it's necessary or not but they can do something like getting a regex from ``{query}`` placeholder and use it in ``rec.re``.

stp-ip:
Still trying to wrap my head around the use case. Can you give a specific example on how {query} + regex would be used.

erbesharat:
URI: https://example.com/path/123/somewhere?regex=(\\d+)
Record: ``"v=txtv0;re={?regex};type=path;code=302"``

This way user can pass the custom regex using query strings too. I don't know if it's going to be even used by anyone or not. Just a use case that came to my mind :smile:

stp-ip:
Something of a interesting use case.
The domain opensourcesoftware.rocks could be used to automatically redirect a subdomain using the "i'm lucky" search from google. Not yet sure how and how that relates to this issue so.

kubernetes.opensourcesoftware.rocks -> https://www.google.com/search?btnI=1&q=kubernetes -> kubernetes.io

```
v=txtv0;to=https://www.google.com/search?btnI=1&q={;type=host;code=302
```

PR #108, opened 2018-09-18
Advanced version for proxy requests

### LINKED ISSUE #86: [proxy] proxy requests instead of redirecting

dbriesz:
Allows for more advanced proxy request configuration such as configuring multiple upstreams and adding headers to the request.

Extends #86

### DIFF
--- a/caddy/caddy.go
+++ b/caddy/caddy.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/mholt/caddy"
 	"github.com/mholt/caddy/caddyhttp/httpserver"
+	"github.com/mholt/caddy/caddyhttp/proxy"
 
 	"github.com/txtdirect/txtdirect"
 	"gopkg.in/natefinch/lumberjack.v2"
@@ -39,6 +40,7 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
 	var enable []string
 	var redirect string
 	var resolver string
+	reverseProxy := proxy.Proxy{}
 
 	c.Next() // skip directive name
 	for c.NextBlock() {
@@ -77,6 +79,17 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
 			}
 			resolver = resolverAddr[0]
 
+		case "proxy":
+			proxyAddr := c.RemainingArgs()
+			if len(proxyAddr) != 1 {
+				return txtdirect.Config{}, c.ArgErr()
+			}
+			upstreams, err := proxy.NewStaticUpstreams(c.Dispenser, proxyAddr[0])
+			if err != nil {
+				return txtdirect.Config{}, err
+			}
+			reverseProxy.Upstreams = upstreams
+
 		case "logfile":
 			logfile := c.RemainingArgs()
 			if len(logfile) != 1 {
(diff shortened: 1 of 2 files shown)


caddy/caddy.go
```diff
@@ -77,6 +79,16 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
 			}
 			resolver = resolverAddr[0]
 
+		case "proxy":
+			toProxy := c.RemainingArgs()
```
stp-ip: Still not sure, what proxy config you thought should work here.

For something like:
```
txtdirect {
  enable proxy
  resolver 127.0.0.1
  proxy / example.com
}
```
The `toProxy` would result in ["/", "example.com"]

caddy/caddy.go
```diff
@@ -77,6 +79,16 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
...
+			if len(toProxy) != 1 {
```
stp-ip: Not sure, why you want to check for this specific length. Could you give a bit more context on what you wanted to do?

caddy/caddy.go
```diff
@@ -77,6 +79,16 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
...
+				return txtdirect.Config{}, c.ArgErr()
+			}
+			upstreams, err := proxy.NewStaticUpstreams(c.Dispenser, toProxy[0])
```
stp-ip: c.Dispenser will error out here as it doesn't get an actual proxy configuration. Let's remove the config logic first and use something hardcoded:
```
upstreams, err := proxy.NewStaticUpstreams(caddyfile.NewDispenser("Testfile", strings.NewReader(`
			proxy / example.com
		`)), "")
```

What that basically does:
* don't set a host (simplification)
* create a new dispenser with the most simple configuration

caddy/caddy.go
dbriesz: I was thinking I could check if the proxy config contains 1 token as a starting point with a simple example. I changed this to return an error if the proxy arguments in the config file are empty instead.

caddy/caddy.go
dbriesz: I added a hardcoded example to start. Here is my local config that I used to test it:

In the txt record:
```
_redirect.two.one     60 IN TXT     "v=txtv0;to=http://google.com;type=proxy;code=302"
```

In the caddyfile:
```
example.test:8080 {
  tls off
  txtdirect {
    enable path host proxy
    resolver 127.0.0.1:5353
    proxy / google.com
    logfile stdout
  }
  log / stdout "{remote} - [{when_iso}] \"{method} {uri} {proto}\" {status} {size} {latency}"
  errors stdout
}
```

stp-ip:
For now I would close this and reiterate, when we have some better ideas on how this should work.

PR #117, opened 2018-09-25
Add documentation for running and testing TXTDirect locally

dbriesz:
Adds documentation for running and testing TXTDirect locally on both Linux and Mac operating systems.

### DIFF
--- a/docs/testing/README.md
+++ b/docs/testing/README.md
@@ -1,0 +1,123 @@
+# Local TXTDirect Configuration
+## Linux
+* Install CoreDNS
+    ```
+    go get github.com/coredns/coredns
+    ```
+* Add testing domain to /etc/hosts e.g. 127.0.0.1 example.test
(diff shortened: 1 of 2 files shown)


docs/testing/README.md
```diff
@@ -0,0 +1,123 @@
...
+    ```
+* Navigate to the directory in terminal where the Corefile was created and start CoreDNS with the Corefile
+    ```
+    coredns -conf Corefile
+    ```
+*  Navigate to the local TXTDirect repository directory in terminal and rebuild TXTDirect
+    ```
+    make build
+    ```
+    *Note: if the above command does not pull in all the changes and create a TXTDirect binary file, use `make travis-build` instead.
```
stp-ip: Good for now, but we should probably fix the difference in local and travis builds in the long run.

stp-ip:
Wrong PR for the comment and I assume that's not the actual config you used for testing. Please comment on #108 instead.

PR #118, opened 2018-09-28
Add e2e tests using mock DNS server

### LINKED ISSUE #17: [e2e] Add more tests
Use go tests with mocking to use custom DNS responses instead of relying on public DNS responses.

Tests could mock the DNS responses, use this mocked server via the custom DNS resolver and check the redirects and end address it's redirected to.

It would simplify testing and make local testing more sophisticated.

erbesharat:
Adds e2e tests for Redirect using the mock DNS 

fixes #17 

For now, it only covers ``Redirect`` function, will add functional tests too to cover caddy part. Also is it ok to remove old ``Redirect`` tests?

### DIFF
--- a/txtdirect_test.go
+++ b/txtdirect_test.go
@@ -413,7 +413,7 @@ func RunDNSServer() {
 	}
 }
 
-func TestRedirectTypeHost(t *testing.T) {
+func TestRedirectE2e(t *testing.T) {
 	tests := []struct {
 		url      string
 		txt      string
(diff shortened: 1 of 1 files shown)


txtdirect_test.go
```diff
@@ -407,6 +409,48 @@ func RunDNSServer() {
 	err := server.ListenAndServe()
 	defer server.Shutdown()
 	if err != nil {
-		log.Fatalf("Failed to start server: %s\n ", err.Error())
+		log.Printf("Failed to start server: %s\n ", err.Error())
+	}
+}
+
+func TestRedirectTypeHost(t *testing.T) {
```
stp-ip: Not sure the function name reflects the actual tests being run.

erbesharat: Oops! forgot to rename it, my first idea was to write single tests for each type :rofl:

stp-ip: TestRedirectE2e?

PR #119, opened 2018-10-02
Add e2e tests for config disable/enable

### LINKED ISSUE #109: [e2e] Test config disable/enable actually disabling specific features
feature

**What happened**:
Currently we minimally test our enable/disable parsing code, but not if these options actually gate any features.

e2e configurations are gated.

erbesharat:
Add e2e tests for config disable/enable

fixes #109

### DIFF
--- a/txtdirect_test.go
+++ b/txtdirect_test.go
@@ -454,3 +454,39 @@ func TestRedirectE2e(t *testing.T) {
 		}
 	}
 }
+
+func TestConfigE2e(t *testing.T) {
+	tests := []struct {
+		url    string
+		txt    string
+		enable []string
+	}{
+		{
+			"https://e2e.txtdirect",
+			txts["_redirect.path.txtdirect."],
+			[]string{},
+		},
+		{
+			"https://path.txtdirect/test",
+			txts["_redirect.path.e2e.txtdirect."],
+			[]string{"host"},
+		},
+		{
+			"https://gometa.txtdirect",
+			txts["_redirect.gometa.txtdirect."],
+			[]string{"host"},
+		},
+	}
+	for _, test := range tests {
+		req := httptest.NewRequest("GET", test.url, nil)
+		resp := httptest.NewRecorder()
+		c := Config{
+			Resolver: "127.0.0.1:" + strconv.Itoa(port),
+			Enable:   test.enable,
+		}
+		err := Redirect(resp, req, c)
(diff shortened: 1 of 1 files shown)


txtdirect_test.go
```diff
@@ -454,3 +454,39 @@ func TestRedirectE2e(t *testing.T) {
...
+func TestConfigE2e(t *testing.T) {
+	tests := []struct {
+		url    string
+		txt    string
+		enable []string
+	}{
+		{
+			"https://e2e.txtdirect",
+			txts["_redirect.path.txtdirect."],
+			[]string{},
```
stp-ip: Shouldn't that actually need `path` to be enabled?

txtdirect_test.go
```diff
@@ -454,3 +454,39 @@ func TestRedirectE2e(t *testing.T) {
...
+		},
+		{
+			"https://path.txtdirect/test",
+			txts["_redirect.path.e2e.txtdirect."],
+			[]string{"host"},
+		},
+		{
+			"https://gometa.txtdirect",
+			txts["_redirect.gometa.txtdirect."],
+			[]string{"host"},
```
stp-ip: Shouldn't that be `gometa` being enabled instead of `host`?

txtdirect_test.go
```diff
@@ -454,3 +454,39 @@ func TestRedirectE2e(t *testing.T) {
...
+		enable []string
+	}{
+		{
+			"https://e2e.txtdirect",
+			txts["_redirect.path.txtdirect."],
+			[]string{},
+		},
+		{
+			"https://path.txtdirect/test",
+			txts["_redirect.path.e2e.txtdirect."],
```
stp-ip: I think we need to rework both the urls to be used and the keys for the DNS records. They don't make their intent as clear from first view.

txtdirect_test.go
erbesharat: Since they are negative tests, we should enable wrong options.

txtdirect_test.go
erbesharat: Yeah, I just wrote some random stuff for records and URLs to just test the functionality. we should discuss and chose nice URLs and right txt records.

txtdirect_test.go
stp-ip: Ah true. Sorry totally missed that. Happy to merge and waiting for positive tests in a follow on PR.

PR #120, opened 2018-10-02
Add parsing for {labelN} placeholder

### LINKED ISSUE #104: [placeholder] Support {labelN}
feature

**What happened**:
We currently don't support {labelN}.
https://caddyserver.com/docs/placeholders

Support {labelN}.

dbriesz:
Adds support for parsing {labelN} placeholders since that feature is not currently supported.

fixes #104

### DIFF
--- a/placeholders.go
+++ b/placeholders.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 )
 
@@ -25,6 +26,17 @@ func parsePlaceholders(input string, r *http.Request) string {
 			input = strings.Replace(input, "{host}", r.URL.Host, -1)
 		case "{hostonly}":
 			input = strings.Replace(input, "{hostonly}", r.URL.Hostname(), -1)
+		case "{label":
+			nStr := placeholder[0][6 : len(placeholder[0])-1] // get the integer N in "{labelN}"
+			n, err := strconv.Atoi(nStr)
+			if err != nil || n < 1 {
+				input = strings.Replace(input, placeholder[0], "", -1)
+			}
+			labels := strings.Split(r.Host, ".")
+			if n > len(labels) {
+				input = strings.Replace(input, placeholder[0], "", -1)
+			}
+			input = strings.Replace(input, placeholder[0], labels[n-1], -1)
 		case "{method}":
 			input = strings.Replace(input, "{method}", r.Method, -1)
 		case "{path}":
--- a/placeholders.go
+++ b/placeholders.go
@@ -1,6 +1,7 @@
 package txtdirect
 
 import (
+	"log"
 	"net/http"
 	"net/url"
 	"path"
@@ -51,10 +52,16 @@ func parsePlaceholders(input string, r *http.Request) string {
 			nStr := placeholder[0][6 : len(placeholder[0])-1] // get the integer N in "{labelN}"
 			n, err := strconv.Atoi(nStr)
 			if err != nil || n < 1 {
+				if err == nil {
+					log.Print("{label0} is not supported")
+				} else {
+					log.Print(err)
+				}
 				input = strings.Replace(input, placeholder[0], "", -1)
 			}
 			labels := strings.Split(r.URL.Hostname(), ".")
 			if n > len(labels) {
+				log.Printf("Cannot parse a label greater than %d", len(labels))
 				input = strings.Replace(input, placeholder[0], "", -1)
 			}
 			input = strings.Replace(input, placeholder[0], labels[n-1], -1)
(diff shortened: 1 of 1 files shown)


placeholders.go
```diff
@@ -25,6 +26,17 @@ func parsePlaceholders(input string, r *http.Request) string {
 			input = strings.Replace(input, "{host}", r.URL.Host, -1)
 		case "{hostonly}":
 			input = strings.Replace(input, "{hostonly}", r.URL.Hostname(), -1)
+		case "{label":
```
erbesharat: Since it's a dynamic placeholder, ``case`` is not gonna work here. You can use an if statement and ``strings.HasPrefix`` or something else after the switch case.

dbriesz: Thanks - I've moved that check outside of the switch statement

dbriesz:
I agree that we need more tests for all the placeholders and most could be added in another PR

placeholders.go
```diff
@@ -46,6 +47,18 @@ func parsePlaceholders(input string, r *http.Request) string {
 			}
 			input = strings.Replace(input, "{user}", user, -1)
 		}
+		if strings.HasPrefix(placeholder[0], "{label") {
+			nStr := placeholder[0][6 : len(placeholder[0])-1] // get the integer N in "{labelN}"
+			n, err := strconv.Atoi(nStr)
+			if err != nil || n < 1 {
+				input = strings.Replace(input, placeholder[0], "", -1)
+			}
+			labels := strings.Split(r.Host, ".")
```
stp-ip: Not really a concern here as we don't use this part of the host, but to be consistent I think we don't need the port and could just use `r.URL.Hostname()`

placeholders.go
```diff
@@ -46,6 +47,18 @@ func parsePlaceholders(input string, r *http.Request) string {
 			}
 			input = strings.Replace(input, "{user}", user, -1)
 		}
+		if strings.HasPrefix(placeholder[0], "{label") {
+			nStr := placeholder[0][6 : len(placeholder[0])-1] // get the integer N in "{labelN}"
+			n, err := strconv.Atoi(nStr)
+			if err != nil || n < 1 {
+				input = strings.Replace(input, placeholder[0], "", -1)
```
stp-ip: Why the decision to not error here? {label0} is not supported so we could just fail or?

placeholders.go
```diff
@@ -46,6 +47,18 @@ func parsePlaceholders(input string, r *http.Request) string {
...
+			}
+			labels := strings.Split(r.Host, ".")
+			if n > len(labels) {
+				input = strings.Replace(input, placeholder[0], "", -1)
```
stp-ip: Why the decision to not error?
Not saying it should error here, just want to know more about the thought process.

placeholders.go
dbriesz: In the caddy implementation of labelN, an empty string is returned instead of failing, so my thinking was that we could do that here. Would it be better to `log.Fatal()` here?

stp-ip: "Someone else did" shouldn't always be the base for a decision. Think through the actual use and failure cases. Why is a decision good or not. Don't just copy and paste. Think about what the actual solution should be doing from both a user perspective and a code perspective.

stp-ip: So let's elaborate what the different failure cases are.
When would this hit?
When would a user do this?
Should we allow a value/placeholder that never actually does anything?
Should we diverge from caddy's implementation and actually make `{label0}` aka the top level tld work?
etc.

placeholders.go
stp-ip: This is a bit different than the above issue.
When would we have a higher number `N` than labels?
Would this make sense to fail?
Could it be actually useful? Say if no subdomain of a subdomain is available it could fallback to returning an empty label.

The other option would be to error and let the actual fallback code handle it. In this case I think it comes down to being strict vs being a bit more forgiving.

placeholders.go
stp-ip: The main point in our usage is probably that we have fallback code that would redirect on error.
So with it being replaced with an empty string we won't trigger the fallback code, but maybe we should.

dbriesz: So an error would occur if the `N` in a label can't be converted to an int by `strconv.Atoi()`. The if statement would also hit if the resulting int is less than 1 since caddy doesn't support a `{label0}`. I don't think we should allow a value/placeholder that doesn't do anything and error instead. And I think for now we shouldn't diverge from the caddy implementation.

stp-ip: That would fill the `err`, but wouldn't even result in a full error, just an empty replace.

Yeah I agree. We should return an error so that in both cases aka no int or {label0} would be handled by the fallback redirect code.

placeholders.go
dbriesz: I'm not entirely sure when we'd have a higher number `N` than labels other than maybe a user error in the txt record. I think I need more information about use cases to decide whether it would be better to error or not here.

stp-ip:
cc @erbesharat Good for another review. Form my side it is lgtm.

PR #127, opened 2018-10-16
Normalize path into allowed DNS chars

erbesharat:
Normalizes path into allowed DNS chars

fixes #126 

**Release note**:

```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -16,6 +16,9 @@ var PathRegex = regexp.MustCompile("\\/([A-Za-z0-9-._~!$'()*+,;=:@]+)")
 var FromRegex = regexp.MustCompile("\\/\\$(\\d+)")
 
 func zoneFromPath(host string, path string, rec record) (string, int, error) {
+	if strings.ContainsAny(path, ".") {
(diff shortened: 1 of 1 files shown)


path.go
```diff
@@ -16,6 +16,9 @@ var PathRegex = regexp.MustCompile("\\/([A-Za-z0-9-._~!$'()*+,;=:@]+)")
 var FromRegex = regexp.MustCompile("\\/\\$(\\d+)")
 
 func zoneFromPath(host string, path string, rec record) (string, int, error) {
+	if strings.ContainsAny(path, ".") {
```
stp-ip: How about any other chars such as "?!()/" do we already normalize them?
Also you feel that replacing "." with "-" is the best way to go? We could just remove them from the string as another option.

erbesharat: Will add other chars too. And I think using ``-`` is a better idea than removing them from the string.

stp-ip: Ok merging this now. But let's add tests for special characters in path in a separate PR.

PR #128, opened 2018-10-18
[WIP] Placeholder tests

### LINKED ISSUE #105: [use-case] Enable subdomain based redirects
feature

Redirecting a request by subdomain. As far as I can see this isn't yet supported by default and might need a few changes within type=host.

Example:
kubernetes.example.com -> https://www.google.com/search?btnI=1&q=kubernetes -> kubernetes.io

```
*.example.com IN CNAME txtdirect.io.
_redirect._.example.com IN TXT "v=txtv0;to=https://www.google.com/search?btnI=1&q={label1};type=host;code=302"
```

Happy to discuss ideas and solutions.

dbriesz:
Adds a number of tests that are needed to ensure that placeholder parsing is working correctly. Also separates placeholder tests from the main TXTDirect tests.

fixes #105 

Additional placeholder tests are needed

### DIFF
--- a/placeholders_test.go
+++ b/placeholders_test.go
@@ -78,6 +78,11 @@ func TestParsePlaceholders(t *testing.T) {
 			"example.com/test/querykey=queryvalue&anotherquerykey=anothervalue",
 		},
 		{
+			"example.com/test/{query_escaped}",
+			"https://example.com/test/?querykey=queryvalue&anotherquerykey=anothervalue",
+			"example.com/test/querykey%3Dqueryvalue%26anotherquerykey%3Danothervalue",
+		},
+		{
 			"about.example.com/{label1}",
 			"https://about.example.com",
 			"about.example.com/about",
--- a/placeholders_test.go
+++ b/placeholders_test.go
@@ -38,6 +38,11 @@ func TestParsePlaceholders(t *testing.T) {
 			"example.com/directory/",
 		},
 		{
+			"example.com/directory/{file}",
+			"https://example.com/directory/file.pdf",
+			"example.com/directory/file.pdf",
+		},
+		{
 			"subdomain.example.com/{label1}",
 			"https://subdomain.example.com/subdomain",
 			"subdomain.example.com/subdomain",
--- a/placeholders_test.go
+++ b/placeholders_test.go
@@ -136,6 +136,10 @@ func TestParsePlaceholdersFails(t *testing.T) {
 			"example.com/{label9000}",
 			"https://example.com/test",
 		},
+		{
+			"example.com/{labelxyz}",
+			"https://example.com/test",
+		},
 	}
 	for _, test := range tests {
 		req := httptest.NewRequest("GET", test.requested, nil)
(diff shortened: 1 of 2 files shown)


placeholders_test.go
```diff
@@ -0,0 +1,132 @@
...
+			"https://project.example.com",
+			"example.com/project.example.com",
+		},
+		{
+			"example.com/{method}",
+			"https://example.com",
+			"example.com/GET",
+		},
+		{
+			"example.com{path}",
```
stp-ip: This should probably test a longer or more complicated path.

placeholders_test.go
```diff
@@ -0,0 +1,132 @@
...
+			"https://example.com/directory/test",
+			"example.com/directory/",
+		},
+		{
+			"example.com/directory/{file}",
+			"https://example.com/directory/file.pdf",
+			"example.com/directory/file.pdf",
+		},
+		{
+			"example.com/?test={fragment}",
```
stp-ip: You are mixing `query` string and `fragment`.
Query: ?querykey=queryvalue&anotherquerykey=anothervalue
Fragment: #uri_fragment_usually

More Details:
https://en.wikipedia.org/wiki/Query_string
https://en.wikipedia.org/wiki/Fragment_identifier

placeholders_test.go
```diff
@@ -0,0 +1,132 @@
...
+			"https://example.com/?test=test",
+			"example.com/?test=test/test",
+		},
+		{
+			"example.com{uri}/{>Test}",
+			"https://example.com/?test=test",
+			"example.com/?test=test/test-header",
+		},
+		{
+			"example.com{uri}/{?test}",
```
stp-ip: Probably best to simplify the test. No need to both test {uri} and {?test} at the same time. Same for the above. Test one thing. Especially, when it's table tests, where duplication doesn't add more code, just test values.

placeholders_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func TestParsePlaceholders(t *testing.T) {
+	tests := []struct {
+		url       string
+		requested string
+		expected  string
+	}{
+		{
+			"example.com{uri}",
+			"https://example.com/?test=test",
```
stp-ip: I would use a more advanced {uri} here.

``` suggestion
https://example.com/about/test/file.html?query=string&another=value#even-a-fragment
```

placeholders_test.go
```diff
@@ -0,0 +1,132 @@
...
+			"example.com%2Fpath_escaped",
+		},
+		{
+			"example.com:{port}",
+			"https://example.com:8080",
+			"example.com:8080",
+		},
+		{
+			"subdomain.example.com/{label1}",
+			"https://subdomain.example.com/subdomain",
```
stp-ip: Using "subdomain" also for the uri or path within the request makes the test harder to understand.
Same for the `label` tests below.

placeholders_test.go
dbriesz: I added a more complex path example. Happy to make additional changes if you have a different example in mind.

placeholders_test.go
dbriesz: Sounds good - I changed the {labelN} tests so that they're easier to understand.

stp-ip:
@erbesharat

PR #130, opened 2018-10-23
Move gometa record defaults to gometa.go

### LINKED ISSUE #77: Move configuration defaults to their own file when possible
See Michael's final comment in #44 for example

erbesharat:
Move gometa record defaults to gometa.go

fixes #77


stp-ip:
Test fails due to the default not working as expected.

PR #132, opened 2018-10-28
Add custom regex with groups support

### LINKED ISSUE #98: [path] Custom regex with numbered matches
feature

**Anything else we need to know?**
regex with numbered groups extracting 2 query strings
`re=\\?query=(?P<2>[^&]+)\\&more=(?P<1>[^&]+)`: _redirect.string.more.example.com
Can't get the above to split it into 2 groups in Go. Have it working on https://regexr.com/ so.

?P<1> is used to name the group, which we could use for order.

erbesharat:
Adds custom regex with groups support

fixes #98

### DIFF
--- a/path_test.go
+++ b/path_test.go
@@ -130,7 +130,7 @@ func Test_zoneFromPath(t *testing.T) {
 			"/12345-some-path?query=string&more=stuff",
 			"",
 			"\\?query=(?P<a>[^&]+)\\&more=(?P<b>[^&]+)",
-			"_redirect.string.stuff.example.com",
+			"_redirect.stuff.string.example.com",
 			nil,
 		},
 		{
@@ -138,7 +138,7 @@ func Test_zoneFromPath(t *testing.T) {
 			"/12345-some-path?query=string&more=stuff&test=testing",
 			"",
 			"\\?query=(?P<b>[^&]+)\\&more=(?P<a>[^&]+)\\&test=(?P<c>[^&]+)",
-			"_redirect.stuff.string.testing.example.com",
+			"_redirect.testing.string.stuff.example.com",
 			nil,
 		},
 		{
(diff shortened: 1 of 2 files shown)


path_test.go
```diff
@@ -125,6 +125,14 @@ func Test_zoneFromPath(t *testing.T) {
 			"_redirect.string.example.com",
 			nil,
 		},
+		{
+			"example.com",
+			"/12345-some-path?query=string&more=stuff",
+			"",
+			"\\?query=(?P<a>[^&]+)\\&more=(?P<b>[^&]+)",
```
stp-ip: We should do a more complicated test that actually reorders the matches.

``` suggestion
			"\\?query=(?P<b>[^&]+)\\&more=(?P<a>[^&]+)\\&test=(?P<c>[^&]+)",
```

Additionally we should test for the correct handling of some errors such as using the same ordered match multiple times.

erbesharat: Will add more tests and error handlings.

path_test.go
```diff
@@ -125,6 +125,30 @@ func Test_zoneFromPath(t *testing.T) {
...
+			"_redirect.string.stuff.example.com",
+			nil,
+		},
+		{
+			"example.com",
+			"/12345-some-path?query=string&more=stuff&test=testing",
+			"",
+			"\\?query=(?P<b>[^&]+)\\&more=(?P<a>[^&]+)\\&test=(?P<c>[^&]+)",
+			"_redirect.stuff.string.testing.example.com",
```
stp-ip: Shouldn't the ordering be:
```
"_redirect.testing.string.stuff"
```
So the query matches `string` as `b`, `stuff` as `a` and `testing` as `c`.
Using our usual ordering would mean _redirect.c.b.a.example.com.
Am I missing something?

erbesharat: Forgot to use ``reverse()``. Changed the tests and used reverse in the last commit.

stp-ip:
Merging with my own review only.

PR #135, opened 2018-10-30
Remove {fragment} placeholder support

dbriesz:
Removes {fragment} placeholder support since it shouldn't be sent to the server and is not something the server supports. Also removes references to fragment in the README.


stp-ip:
@erbesharat to bring you up to speed. Usually the fragment of the url is never send to the server so there is not real reason to handle it on the server-side.
What are your thoughts? Have we missed something?

PR #136, opened 2018-10-30
Add documentation regarding how multi-level tlds are handled

### LINKED ISSUE #122: Document how multi-level tlds are handled
feature

**What happened**:
TXTDirect does not currently have documentation regarding how multi-level tlds are handled.

The docs should include an explanation and example about how TXTDirect handles multi-level tlds. Using example.co.uk, "co" would be treated as a subzone from a DNS perspective and would be used as {label1}.

dbriesz:
Adds documentation that explains how multi-level tlds are handled since the current docs do not include that information.

fixes #122

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -45,7 +45,7 @@ External Links:
     https://en.wikipedia.org/wiki/Percent-encoding  
     https://tools.ietf.org/html/rfc3986#page-11
 
-For multi-level tlds such as `example.co.uk`, "co" is treated as a subzone from a DNS perspective and would be used as {label1}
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -45,6 +45,8 @@ External Links:
     https://en.wikipedia.org/wiki/Percent-encoding  
     https://tools.ietf.org/html/rfc3986#page-11
 
+For multi-level tlds such as `example.co.uk`, "co" is treated as a subzone from a DNS perspective and would be used as {label1}
```
stp-ip: {label1} would be `example.com` in this case or?
{label2} would be `co` and {label3} would be `uk`

dbriesz: Right, I changed the wording to accurately reflect how this example would be handled. I also changed the inline comment.

PR #141, opened 2018-10-31
Implement gomod type

erbesharat:
Implements gomod type

**Release note**:

```release-note

```


txtdirect.go
```diff
@@ -336,7 +341,7 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 		if err != nil {
 			return err
 		}
-		reverseProxy := proxy.NewSingleHostReverseProxy(u, "", proxyKeepalive, proxyTimeout, fallbackDelay)
+		reverseProxy := proxy.NewSingleHostReverseProxy(u, "", proxyKeepalive, proxyTimeout)
```
stp-ip: This breaks with the latest Caddy master branch.

erbesharat: I thought I had discarded this change. Will bring it back.

txtdirect.go
```diff
@@ -337,7 +337,7 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 		if err != nil {
 			return err
 		}
-		reverseProxy := proxy.NewSingleHostReverseProxy(u, "", proxyKeepalive, proxyTimeout)
+		reverseProxy := proxy.NewSingleHostReverseProxy(u, "", proxyKeepalive, proxyTimeout, fallbackDelay)
```
erbesharat: This line returned a ``too many arguments`` error at first for me so I've removed the fallback argument to make it work.
It happened because my local caddy package was not updated, they have added it in mholt/caddy#2309

erbesharat:
Tests need to get refactored and there is also an edge-case if the version is not just a simple semantic version the ``modVersionRegex`` will not extract the whole version from the path.
Example:
```
github.com/erbesharat/gogenius/@v/v0.0.0.mod # works
github.com/erbesharat/gogenius/@v/v0.0.0-20171219132314-7bdf14fa257c.mod  # fails
```

erbesharat:
The build for version ``v.1.10.x`` fails because gomods is not supported on that version. 
Should we remove it from Travis's go versions?

alarsyo:
I guess 1.11 is old enough to only keep that one now

erbesharat:
Changed the config parser and removed the nested switch cases. Ready for a final review.

stp-ip:
Taking some time out tomorrow and tuesday for another final review and test session. \o/

erbesharat:
Great, btw since config parser got changed in this PR, will wait for it to get merged then will change Prometheus's config (170).

gomods.go
```diff
@@ -0,0 +1,227 @@
...
+		if err != nil {
+			return c.ArgErr()
+		}
+		gomods.Enable = value
+	case "gobinary":
+		if value == "" {
+			value = os.Getenv("GOROOT") + "/bin/go"
+		}
+		gomods.GoBinary = value
+	case "workers":
```
stp-ip: If workers is not specified it should default to either `1` or use the number of cores available. Probably a conservative `1` is the better solution.

gomods.go
```diff
@@ -0,0 +1,227 @@
...
+	switch value {
+	case "enable":
+		value, err := strconv.ParseBool(value)
+		if err != nil {
+			return c.ArgErr()
+		}
+		gomods.Enable = value
+	case "gobinary":
+		if value == "" {
+			value = os.Getenv("GOROOT") + "/bin/go"
```
stp-ip: Setting a default value is good, but this would only be the default, when `gobinary` is added to the config. So a general default seems better.

gomods_test.go
```diff
@@ -0,0 +1,51 @@
...
+			path: "/github.com/okkur/reposeed-server/@v/v0.1.0.zip",
+		},
+	}
+	for _, test := range tests {
+		if err := os.MkdirAll("/tmp/gomods", os.ModePerm); err != nil {
+			t.Fatal("Couldn't create storage directory (/tmp/gomods)")
+		}
+		w := httptest.NewRecorder()
+		r := httptest.NewRequest("GET", fmt.Sprintf("https://example.com%s", test.path), nil)
+		c := Config{
```
stp-ip: Needs more config specific tests. Also around the defaulting mentioned before.

gomods.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+		gomods.GoBinary = value
+	case "workers":
+		value, err := strconv.Atoi(value)
+		if err != nil {
+			return c.ArgErr()
+		}
+		gomods.Workers = value
+	case "cache":
```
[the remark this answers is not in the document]
stp-ip: Maybe even better would be using an actual temporary directory provided by afero as default. 
That could be called `cache.type = "tmp"`
That would leave `tmp` not taking a path as it's provided by afero.
And `local`, which would need a path set.

The next iteration then would be a object store being supported such as GCS to have a more scalable system and a higher hit rate. But that's for a follow on pr.

.travis.yml
```diff
@@ -12,6 +12,9 @@ script:
   - go get -u github.com/mholt/caddy
   - go get -u gopkg.in/natefinch/lumberjack.v2
   - go get -u github.com/miekg/dns
+  - go get -u github.com/gomods/athens/...
+  - rm -rf ${TRAVIS_HOME}/gopath/src/github.com/gomods/athens/vendor/github.com/spf13/afero
```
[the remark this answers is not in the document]
erbesharat: Because it tries to use the one from athens's vendor inside our project and we end up with an error like this:
```
./gomods.go:125:49: cannot use fs (type "github.com/spf13/afero".Fs) as type "github.com/gomods/athens/vendor/github.com/spf13/afero".Fs in argument to module.NewGoGetFetcher:
```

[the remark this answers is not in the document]
erbesharat: Couldn't find anything or maybe used the wrong keywords for search. Handled it the ugly way for now till we find a better solution.

gomods.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+type ModuleHandler interface {
+	fetch(r *http.Request, c Config) (*storage.Version, error)
+	storage(c Config) (storage.Backend, error)
+	dp(fetcher module.Fetcher, s storage.Backend, fs afero.Fs) download.Protocol
+}
+
+var gomodsRegex = regexp.MustCompile("(list|info|mod|zip)")
+var modVersionRegex = regexp.MustCompile("@v\\/(v\\d+\\.\\d+\\.\\d+\\-\\d+\\-[\\w\\d]+|v\\d+\\.\\d+\\.\\d+|v\\d+\\.\\d+|latest|master)")
```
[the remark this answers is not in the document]
stp-ip: As discussed. Let's do that in a follow on PR after more testing on my side. Happy for you to ping me on it.

gomods.go
```diff
@@ -0,0 +1,253 @@
...
+	s, err := m.storage(c)
+	if err != nil {
+		return nil, err
+	}
+	dp := m.dp(fetcher, s, c)
+	return dp, nil
+}
+
+func (m Module) storage(c Config) (storage.Backend, error) {
+	switch c.Gomods.Cache.Type {
```
stp-ip: Shouldn't "tmp" be handled here as well? Or am I missing something?

stp-ip: Ok we are setting the tmp path/fs within the defaulting section. This feels a bit weird, but I see the reasoning.

stp-ip:
\o/

PR #147, opened 2018-11-17
Add a query for _redirect._ if _redirect isn't found

### LINKED ISSUE #145: Handle missing records in case of DNS wildcards via additional TXT record
feature

**What happened**:
Currently we always try to query `_redirect` as zone for a given request. With the subdomain use case outlined in #105 this is not easily doable currently.
We can't set a wildcard _redirect record and other workarounds are not that feasible. 

Query `_redirect._`, if `_redirect` isn't available.
Say _redirect.wildcard.example.com is not available as we can't possibly set a wildcard such as `_redirect.*`.

Not yet sure, if that impacts us with additional query load.
Probably depends on the level it makes sense to implement it.
My assumption would be, if we are on the type=path code flow already, we don't need this as we already have a wildcard approach.
With any other request (basically all primary requests), if _redirect isn't found we should additionally check _redirect._ . This way we can circumvent the wildcard issue and don't add more load to any other requests besides this edge case.
Happy to discuss this especially, if you have ideas on problems or impact this could have.

dbriesz:
This PR adds a query for _redirect._ if the _redirect query isn't found for primary requests. This solves the wildcard issue where we can't set a wildcard _redirect record.

fixes #145

### DIFF
--- a/placeholders.go
+++ b/placeholders.go
@@ -26,9 +26,9 @@ func parsePlaceholders(input string, r *http.Request) (string, error) {
 		case "{fragment}":
 			input = strings.Replace(input, "{fragment}", r.URL.Fragment, -1)
 		case "{host}":
-			input = strings.Replace(input, "{host}", r.URL.Host, -1)
+			input = strings.Replace(input, "{host}", r.Host, -1)
 		case "{hostonly}":
-			input = strings.Replace(input, "{hostonly}", r.URL.Hostname(), -1)
+			input = strings.Replace(input, "{hostonly}", r.Host, -1)
 		case "{method}":
 			input = strings.Replace(input, "{method}", r.Method, -1)
 		case "{path}":
@@ -51,7 +51,7 @@ func parsePlaceholders(input string, r *http.Request) (string, error) {
 			input = strings.Replace(input, "{user}", user, -1)
 		}
 		/* For multi-level tlds such as "example.co.uk", "co" would be used as {label2},
-		 "example" would be {label1} and "uk" would be {label3} */
+		"example" would be {label1} and "uk" would be {label3} */
 		if strings.HasPrefix(placeholder[0], "{label") {
 			nStr := placeholder[0][6 : len(placeholder[0])-1] // get the integer N in "{labelN}"
 			n, err := strconv.Atoi(nStr)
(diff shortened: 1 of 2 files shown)


stp-ip:
Thinking through different possibilities, if the current idea doesn't work with the DNS spec.

placeholders.go
```diff
@@ -26,9 +26,9 @@ func parsePlaceholders(input string, r *http.Request) (string, error) {
 		case "{fragment}":
 			input = strings.Replace(input, "{fragment}", r.URL.Fragment, -1)
 		case "{host}":
-			input = strings.Replace(input, "{host}", r.URL.Host, -1)
+			input = strings.Replace(input, "{host}", r.Host, -1)
 		case "{hostonly}":
-			input = strings.Replace(input, "{hostonly}", r.URL.Hostname(), -1)
+			input = strings.Replace(input, "{hostonly}", r.Host, -1)
```
stp-ip: This one needs the port cleanup code as we take the plain r.Host, which still can have a port.

placeholders.go
```diff
@@ -61,7 +61,7 @@ func parsePlaceholders(input string, r *http.Request) (string, error) {
 			if n < 1 {
 				return "", fmt.Errorf("{label0} is not supported")
 			}
-			labels := strings.Split(r.URL.Hostname(), ".")
+			labels := strings.Split(r.Host, ".")
```
stp-ip: r.Host also needs the port code as we use the request's Host, which still can have a port.

txtdirect.go
```diff
@@ -165,14 +165,31 @@ func contains(array []string, word string) bool {
...
+		if txts[0] == "" {
+			// if record empty, jump into wildcards again
+			hostSlice := strings.Split(host, ".")
+			hostSlice[0] = "_"
+			host = strings.Join(hostSlice, ".")
+			txts, err = query(host, ctx, c)
+			if err != nil {
+				return record{}, err
+			}
+		}
```
stp-ip: This is duplicated. First we are in the error check. That should still be separated.
And with the above code, if we have an error an empty record is implied.

txtdirect.go
```diff
@@ -165,14 +165,24 @@ func contains(array []string, word string) bool {
...
+		if err != nil || txts[0] == "" {
+			// if error present or record empty, jump into wildcards
+			hostSlice := strings.Split(host, ".")
+			hostSlice[0] = "_"
+			host = strings.Join(hostSlice, ".")
+			txts, err = query(host, ctx, c)
+		}
+		if err != nil {
+			return record{}, err
+		}
```
stp-ip:
```
	txts, err := query(host, ctx, c)
	if err != nil {
		log.Printf("Initial DNS query failed: %s", err)
	}

	if err != nil || txts[0] == "" {
		// if error present or record empty, jump into wildcards
		hostSlice := strings.Split(host, ".")
		hostSlice[0] = "_"
		host = strings.Join(hostSlice, ".")
		txts, err = query(host, ctx, c)
		if err != nil {
			return record{}, err
		}
	}
```

From a first glance I would say this is cleaner.

PR #150, opened 2018-11-26
[docerv2]: Add error handling

### LINKED ISSUE #149: [dockerv2] Add fallback path
feature

**What happened**:
Currently we don't have good error handling and don't use fallback.

When accessing the root it should redirect to the configured root. If requests don't match docker it should also go into the usual fallback mechanism.

erbesharat:
Adds error handling for dockerv2

fixes #149 

**Release note**:

```release-note

```

### DIFF
--- a/dockerv2.go
+++ b/dockerv2.go
@@ -1,10 +1,12 @@
 package txtdirect
 
 import (
+	"log"
 	"net/http"
 	"net/url"
 	"regexp"
 	"strings"
+	"time"
 )
 
 var dockerRegexes = map[string]*regexp.Regexp{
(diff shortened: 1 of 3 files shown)


dockerv2.go
```diff
@@ -14,6 +16,10 @@ var dockerRegexes = map[string]*regexp.Regexp{
 
 func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error {
 	path := r.URL.Path
+	if !strings.Contains(path, "/v2") {
```
stp-ip: HasPrefix seems more appropriate or?

dockerv2.go
```diff
@@ -14,6 +16,10 @@ var dockerRegexes = map[string]*regexp.Regexp{
...
+		log.Printf("<%s> [txtdirect]: unrecognized path for dockerv2: %s", time.Now().Format(logFormat), path)
+		fallback(w, r, rec.Root, http.StatusPermanentRedirect, Config{})
```
stp-ip: Do we agree that we should use the fallback mechanism here and if rec.Root is provided always redirect there?
This is different than the other fallbacks. Also it seemed you disliked the idea of rec.Root being used as a general fallback.
Other options is use an additional fallback value, redirect to rec.To (doesn't perfectly work with base redirects such as "gcr.io/".

/ -> rec.Root
/ -> !rec.Root -> fallback
/random -> ?

PR #160, opened 2018-12-24
Add go-source meta tag support

### LINKED ISSUE #142: Support go source meta tag withing gometa/gomods
feature

**What happened**:
Currently vanity urls done by gometa or gomods don't support godoc via the go-source meta tag.

Add go-source meta tag.

erbesharat:
Adds go-source meta tag support

fixes #142 

**Release note**:

```release-note

```

### DIFF
--- a/gometa.go
+++ b/gometa.go
@@ -24,7 +24,7 @@ var tmpl = template.Must(template.New("").Parse(`<!DOCTYPE html>
 <html>
 <head>
 <meta name="go-import" content="{{.Host}}{{.Path}} {{.Vcs}} {{.NewURL}}">
-<meta name="go-source" content="{{.Host}}{{.Path}} _ {{.NewURL}}/tree/v2{/dir} {{.NewURL}}/blob/v2{/dir}/{file}#L{line}">
+{{if .HasGoSource}}<meta name="go-source" content="{{.Host}}{{.Path}} _ {{.NewURL}}/tree/v2{/dir} {{.NewURL}}/blob/v2{/dir}/{file}#L{line}">{{end}}
 </head>
 </html>`))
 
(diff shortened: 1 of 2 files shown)


gometa_test.go
```diff
@@ -48,6 +49,7 @@ func TestGometa(t *testing.T) {
 <html>
 <head>
 <meta name="go-import" content="empty.com/test git ">
+<meta name="go-source" content="empty.com/test _ /tree/v2{/dir} /blob/v2{/dir}/{file}#L{line}">
```
erbesharat: Still not sure what we should do in the ``go-source`` tag when ``rec.To`` is empty.

gometa.go
```diff
@@ -24,6 +24,7 @@ var tmpl = template.Must(template.New("").Parse(`<!DOCTYPE html>
 <html>
 <head>
 <meta name="go-import" content="{{.Host}}{{.Path}} {{.Vcs}} {{.NewURL}}">
+<meta name="go-source" content="{{.Host}}{{.Path}} _ {{.NewURL}}/tree/v2{/dir} {{.NewURL}}/blob/v2{/dir}/{file}#L{line}">
```
stp-ip: why "v2"? couldn't it also be "master" or any other tag?
This feels too hardcoded to be workable on all various code platforms.

Any documentation you used to make that decision? Maybe I'm just missing some information.

Also does the #L syntax work on all code sites? Also where does the line come from?

gometa_test.go
stp-ip: What are the cases, where rec.To is empty?
Maybe not expose "go-source" at all?

gometa.go
erbesharat: I used the examples from https://github.com/golang/gddo/wiki/Source-Code-Links
Couldn't find any other documentation about it.

stp-ip: The example uses a specific version tag instead of master. That's a special case in my view.
Also it seems too specific on Github. Will check, if I can find a bit more info on it.

stp-ip: Couldn't find anything either. Will keep looking. Happy for you to take another look. It should support the repositories that go-import supports and probably reference master or a specific version set in our URL?

erbesharat: Searched a lot but couldn't find any documentation about it. need to check the code itself.

stp-ip: Yeah seems like quite a non documented feature. Was looking into supporting domains, but couldn't fine much either.

alarsyo:
What's the status on this ? Can I help ?

Looks like the main problem is multi site compatibility, so why not:

1. Let the user customize the format of the go source, depending on what hosting website he uses himself
2. Eventually, if no custom format is provided, try to detect a known website (github, gitlab...)

PR #163, opened 2019-01-02
Add regex matches as placeholders

### LINKED ISSUE #148: [placeholder] Add regex matches as placeholders
feature

**What happened**:
Currently we can use regex matches to choose the used TXT record.

Use match as placeholder within a specific or wildcard record.
{$1} or {$one} as syntax?

erbesharat:
Adds regex matches as placeholders

fixes #148 

I've passed ``pathSlice`` as a pointer to zoneFromPath because I didn't want to return more than 3 variables. If it's ok to be more than 3, can change it.

**Release note**:

```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -20,7 +20,7 @@ var GroupOrderRegex = regexp.MustCompile("P<([a-zA-Z]+[a-zA-Z0-9]*)>")
 // zoneFromPath generates a DNS zone with the given host and path
 // It will use custom regex to parse the path if it's provided in
 // the given record.
-func zoneFromPath(host string, path string, rec record) (string, int, error) {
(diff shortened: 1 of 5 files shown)


stp-ip:
/LGTM

I'm neutral on the pointer. It makes the function a bit harder to understand, but let's wait for @alarsyo on this too.

erbesharat:
I agree that having a pointer is more ugly and complex than 4 return values. @stp-ip What do you think? Should we go with returning 4 values? As Antoine mentioned, that way we can avoid passing empty slices to function too.

stp-ip:
Also here: Let's use 4 and refactor later.

Issue #175, opened 2019-01-13
[prometheus] Status shown as status="ĭ"


stp-ip:
**Is this a BUG REPORT or FEATURE REQUEST?**:
bug

This feels like a status bug. Not sure where it comes from. Didn't yet investigate, but wanted to log it for now.

Issue #173, opened 2019-01-13
[prometheus] Rename current metric names


stp-ip:
**Is this a BUG REPORT or FEATURE REQUEST?**:
feature

**What happened**:
We currently don't adhere to the normal prometheus naming recommendations.

**What you expected to happen**:
Use convention. https://github.com/prometheus/docs/blob/master/content/docs/practices/naming.md
https://prometheus.io/docs/instrumenting/writing_exporters/

**Anything else we need to know?**:
txtdirect_total_requests_per_host{host}
->
txtdirect_redirect_count_total{host}

txtdirect_total_returned_statuses_per_host{host,status}
->
txtdirect_redirect_status_count_total{host,status}

Issue #172, opened 2019-01-13
[prometheus] Add txtdirect_fallback_type_count_total{host, type}


stp-ip:
**Is this a BUG REPORT or FEATURE REQUEST?**:
feature

**What happened**:
Currently we have no idea how often fallbacks are being triggered.

**What you expected to happen**:
Add metrics about what specific fallback types are triggered.

stp-ip:
**What happened**:
We currently don't have any stats on fallback usage.

**What you expected to happen**:
Add counters for fallback usage.

**Anything else we need to know?**:
Could add "fallback=$see-below".
The type here should be something along the lines of:
 * "redirect" or "global", when the config setting is used
 * "wildcard", when using _redirect.\_.example.test after _redirect.example.test isn't available
 * "root", when using rec.Root
 * "to", when using rec.To from path record
 * "website", when using website in types such as dockerv2

Examples:
{host=example.test,type=host,fallback=global}
{host=example.test,type=proxy,fallback=global}
{host=example.test,type=host,fallback=wildcard}
{host=example.test,type=host,fallback=root}
{host=example.test,type=path,fallback=root}
{host=example.test,type=path,fallback=to}
{host=example.test,type=dockerv2,fallback=website}

PR #178, opened 2019-01-14
[prometheus]: Add more metrics and rename current ones 

### LINKED ISSUE #171: [prometheus] Add txtdirect_redirect_type_count_total{host,type}
feature

**What happened**:
Currently we don't have any insight into how often types are used.

Add a counter that counts up each usage of a type.

So during one request it counts up the usage and marks multiple with multiple labels corresponding to the type.

So path + host would be txtdirect_redirect_type_count_total{host=example.com,type=path,type=host}=1
Host would be txtdirect_redirect_type_count_total{host=example.com,type=host}=1

erbesharat:
Adds more metrics and rename current ones

fixes #171, #172, #173, #175

### DIFF
--- a/caddy/caddy.go
+++ b/caddy/caddy.go
@@ -196,6 +196,14 @@ func (rd Redirect) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error
 		}
 		return http.StatusInternalServerError, err
 	}
+
+	// Count total redirects if prometheus is enabled
+	if w.Header().Get("Status-Code") == "301" || w.Header().Get("Status-Code") == "302" {
+		if rd.Config.Prometheus.Enable {
+			txtdirect.RequestsCount.WithLabelValues(r.Host).Add(1)
+		}
+	}
+
 	return 0, nil
 }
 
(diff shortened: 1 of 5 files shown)


stp-ip:
This needs a bit more testing. Especially around the fallback change.

stp-ip:
During testing the stats didn't seem to be correct. I can dive in a bit more, but a few things already:

On a 500 error:
* txtdirect_redirect_count_total counts twice
* txtdirect_redirect_status_count_total doesn't count 500, but 200

Using type=gometa:
* txtdirect_redirect_count_total counts twice
* txtdirect_redirect_status_count_total counts 200 and 302 (when code is set to 302)

Using type=path:
* wildcard flow isn't counted (_redirect._.example.test)
* fallback flow is missing type (_redirect.example.test using rec.To) <- the type here should be something along the lines of "redirect" or "global" (aka the config setting), "wildcard", "root", "fallback" (aka rec.To used in type=path? better name?), "website" (aka website within dockerv2 for example) <- happy to discuss this more

stp-ip:
Also favicon.ico 404 is counted as 200 status.

stp-ip:
Move out the fallback fix into it's own issue #199 . And currently updating the issues with this PR.

stp-ip:
On the double counting issue.
I think we should count a redirect once for txtdirect_redirect_status_count_total, but each usage of txtdirect_redirect_type_count_total?

So a redirect that uses path + host would have count_total +1, but also type_count_total +2 (host +1 and path +1). What do you think?

PR #180, opened 2019-01-29
Refactor Makefile and Dockerfile

erbesharat:
Refactor Makefile and Dockerfile

Remaining:

- Add the project's path dynamically to the Dockerfile (Line 4 and 6)
- Add push target to the makefile
- ...

I will add whatever comes to my mind to the list above, let me know if we need to add something else or change something.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,14 @@
-FROM alpine:3.6
-RUN apk --no-cache add ca-certificates
-ADD txtdirect /caddy
-CMD ["/caddy"]
+FROM golang:1.11-alpine
+
+RUN apk --no-cache add ca-certificates make git && \
+  mkdir -p $GOPATH/src/github.com/txtdirect/txtdirect
+
+WORKDIR $GOPATH/src/github.com/txtdirect/txtdirect
+
+COPY . .
+
+RUN make dependencies
+
+RUN make recipe
+
+CMD ["./txtdirect"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,6 @@ FROM alpine:3.9
 
 RUN apk --no-cache add ca-certificates
 
-ADD txtdirect /caddy
+ADD txtdirect /txtdirect
 
-CMD ["/caddy"]
+CMD ["/txtdirect"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,7 @@
-FROM golang:1.11-alpine
+FROM alpine:3.9
 
-RUN apk --no-cache add ca-certificates make git && \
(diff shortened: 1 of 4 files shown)


stp-ip:
So a few things already:
* currently this means we are shipping the golang build container within the txtdirect container, which seems overhead, that is unnecessary
* building a alpine based image, with just the runtime needed assets (for now the binary, later the binary + tor) makes this easier
* why use a build container with a hardcoded docker file? Couldn't we pass in the go build command to a build container and mount in the working directory and then pull in deps within the container and build it?
* same idea could be used for testing, but this might make testing locally a bit harder as docker build be needed.

As long as in both cases we use the make file as source of truth it doesn't really matter.

Travis:
"docker run -v $(pwd):/src golang:1.11-alpine /bin/sh -c "make test" <- testing stage (uses make dep under the hood)
"docker run -v $(pwd):/src golang:1.11-alpine /bin/sh -c "make deploy" <- deploy stage (uses make dep and make build under the hood)

Dev:
make test and/or make build

Happy to iterate with your thoughts.

stp-ip:
The other option is to require docker being installed to run the build. This gives us the ability to build with the latest image we provide and don't have to care much about the users install. As we are not going with using the full build template, it seems the middle ground with having make as the source of truth and using docker for isolation within travis is ok.

stp-ip:
Adding to this. Using docker for both the ci and the user makes the overall interaction with changed code less likely. A test will always run with fresh deps so switching between projects or changing something locally never is an issue as local deps are never relied upon. This requires more resources/bandwidth overall so I understand that this might not be perfect, which brings us back to the middle ground, where we have the option to built and test with local deps, but it's also easy before the final push to do a clean run inside docker and the same can be used within travis.

So maybe have the option to use docker for any command, which just mounts the sourcecode and runs the make command within a container.
make docker test -> docker run .... make test

Dockerfile
```diff
@@ -1,14 +1,7 @@
-FROM golang:1.11-alpine
+FROM alpine:3.9
 
-RUN apk --no-cache add ca-certificates make git && \
-  mkdir -p $GOPATH/src/github.com/txtdirect/txtdirect
+RUN apk --no-cache add ca-certificates
 
-WORKDIR $GOPATH/src/github.com/txtdirect/txtdirect
+ADD txtdirect /caddy
```
stp-ip: Why add it as `/caddy`?

erbesharat: It was `/caddy` in the previous version, so I thought maybe there is a reason for it :smile:

stp-ip: Not that I know of. Might just have been an oversight from before :)

PR #195, opened 2019-02-04
Fix Makefile bugs

### LINKED ISSUE #194: [make] Iterate on makefile refactor
* [x] `make test`: Runs test on caddy-copy code, if `make build` was previously run.
* [x] `make build`: Failing build, if caddy-copy is already present. We should either clean it up or better only update the caddy-copy code.
* [x] `make image-build`: Fails without previous build run.
* [x] We don't do a full test run before deploying/pushing the container to GCR. One could have failing tests, but still build and deploy it.
* [x] `make docker-test` and `make docker-build` mount current directory and have issues, when `make build` or `make test` or `make docker-test` or `make docker-build` was run before aka when caddy-copy is still present.

What was your general routine to not run into the above issues? @erbesharat

erbesharat:
Fixes Makefile bugs

fixes #194 

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ CONTAINER ?= $(BIN)
 .DEFAULT_GOAL := build
 
 recipe:
-	git clone https://github.com/mholt/caddy caddy-copy
+	if [ -d caddy-copy ]; then cd caddy-copy && git pull && cd ..; else git clone https://github.com/mholt/caddy caddy-copy; fi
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -41,6 +41,7 @@ test: dependencies
 	go test -v `go list ./... | grep -v caddy-copy`
 
 image-build:
+	if [ ! -f $(BIN) ]; then make docker-build; fi
```
stp-ip: We might run the image-build step anyway or else we might end up with an older binary in the container image or?

PR #197, opened 2019-02-06
Trigger fallback when the request's host is an IP

### LINKED ISSUE #190: [redirect] Redirect fallback not triggered with "127.0.0.1"
bug

**What happened**:
With "redirect" config set and using "127.0.0.1", the dns queries fail, which results in a 500 error.

Redirect to the redirect config instead.

erbesharat:
Triggers fallback when the request is for 127.0.0.1

fixes #190

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -279,6 +279,15 @@ func query(zone string, ctx context.Context, c Config) ([]string, error) {
 	return txts, nil
 }
 
+func isIP(host string) bool {
+	hostSlice := strings.Split(host, ".")
+	_, err := strconv.Atoi(hostSlice[len(hostSlice)-1])
+	if err == nil {
+		return true
+	}
+	return false
+}
+
 // Redirect the request depending on the redirect record found
 func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 	host := r.Host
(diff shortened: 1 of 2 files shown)


txtdirect.go
```diff
@@ -301,6 +301,12 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 		return nil
 	}
 
+	if host == "127.0.0.1" {
```
stp-ip: That only partially solves the general problem.
What about any other used IP etc.?
We should handle IPs correctly either way so that it always either goes right back to the instance redirect or 404. What do you think?

txtdirect.go
```diff
@@ -279,6 +279,15 @@ func query(zone string, ctx context.Context, c Config) ([]string, error) {
 	return txts, nil
 }
 
+func isIP(host string) bool {
+	hostSlice := strings.Split(host, ".")
+	_, err := strconv.Atoi(hostSlice[len(hostSlice)-1])
```
stp-ip: That's an interesting way of checking for an IP.
Not saying it's incorrect, but this doesn't cover IPv6.

So let's add IPv6 within the tests and check, if this still holds true.
Maybe there is also an underlying issue with the fallback not being triggered for IPs even so it should have.

stp-ip: What I mean. I think the fallback should have triggered due to the failing lookups already so it might be a bug in the code and not that specific to IPv4 or IPv6 addresses.

txtdirect.go
```diff
@@ -279,6 +279,15 @@ func query(zone string, ctx context.Context, c Config) ([]string, error) {
...
+	if err == nil {
+		return true
+	}
+	return false
```
alarsyo: This could just be 

```go
return err == nil
```

PR #213, opened 2019-02-12
[test]: Add e2e fallback tests

### LINKED ISSUE #206: [tests] We need better fallback tests
feature

**What happened**:
We currently run into issues with fallback too often.

Test the fallback path for each type.

erbesharat:
Adds e2e fallback tests

fixes #206


erbesharat:
I think for simplifying the main test file (txtdirect_test.go) is to move all the other tests to their own file and only keep redirect tests + the mock zonefile/zonemap inside the txtdirect_test.go
It doesn't make that much of change but at least we can move for example the record parsing code and its tests to another file to keep the main test file a little cleaner.

I still don't have any idea on how to make e2e tests and the zonemap more simple. Need more time to think about it. Also, think it would be good to create another issue for e2e tests refactor and work on it in another PR.

PR #216, opened 2019-02-16
[logfile]: Fix empty value panic and add tests

### LINKED ISSUE #215: [logfile] empty logfile config panics
bug

**What happened**:
Setting logfile without anything panics.
```
txtdirect {
  enable host
  logfile
}
```

Don't panic.

Is this intended? What should we use as default?
Might be worth writing a test for.

erbesharat:
Fixes empty value panic and add tests

fixes #215

### DIFF
--- a/caddy/caddy.go
+++ b/caddy/caddy.go
@@ -82,11 +82,9 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
 
 		case "logfile":
 			// Set stdout as the default value
-			if !c.NextArg() {
-				parseLogfile(logfile)
-				continue
+			if c.NextArg() {
+				logfile = c.Val()
 			}
-			logfile = c.Val()
 			parseLogfile(logfile)
 		case "gomods":
 			gomods.Enable = true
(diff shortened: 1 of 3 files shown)


caddy/caddy.go
```diff
@@ -80,11 +81,13 @@ func parse(c *caddy.Controller) (txtdirect.Config, error) {
...
-			if len(logfile) != 1 {
-
+			// Set stdout as the default value
+			if !c.NextArg() {
+				parseLogfile(logfile)
+				continue
 			}
-			parseLogfile(logfile[0])
+			logfile = c.Val()
+			parseLogfile(logfile)
```
alarsyo: I feel like rewriting this part like this would be clearer:

```go
if c.NextArg() {
    logfile = c.Val()
}
parseLogfile(logfile)
```

stp-ip:
I think defaulting to stdout, when logfile is specified is good. But defaulting to "logfile stdout", if nothing is specified will log quite a bit of output. Currently this log is mostly debug output and prints multiple lines per request. Having the full log enabled by default seems too much. What do you think?

stp-ip:
Maybe I'm missing something, but that still enables logging by default and by default it also logs to stdout right?

erbesharat:
@stp-ip It doesn't log by default. `logfile` variable is set to "stdout" but `parseLogfile` function only gets called when `logfile` field is available in the config.

stp-ip:
You are right. Merging ^^

PR #223, opened 2019-02-24
[fallback]: Fix missing argument in the function call

erbesharat:
Fixes missing argument in the function call.

P.S: This bug causes master tests to fail.


stp-ip:
This should have come up in my review. Seems like I missed this. Sorry.

PR #224, opened 2019-02-25
Add tests and refactor a few of the old tests

erbesharat:
Adds tests and refactors a few of the old tests

Added tests for all of the functions inside the `txtdirect.go`. Let me know if I've missed anything. Maybe some of the tests look unnecessary but I believe we should have tests for everything, even if a function only returns a predefined string. But anyway, let me know if something seems unnecessary and we can disscuss it.
Also about the e2e tests, I only removed a field from the struct that wasn't needed. Couldn't think of any way to make it more simple, let me know if you have any suggestions or ideas.

P.S: We can change this to a WIP PR and work on tests for other parts of the project too.
P.S1: Think it would be good to add unit tests for config parsers too.
P.S2: We should discuss what parts of the project need negative tests or we can even add negative tests for all of the parts too.


stp-ip:
As a follow on I think we can also remove the test struct for fallbackE2E as you did with the redirectE2E tests.

PR #225, opened 2019-02-25
[make]: Add caddy-prometheus middleware as the first middleware

### LINKED ISSUE #169: [make] Add caddy-prometheus plugin/middleware as the first plugin
feature

**What happened**:
Currently prometheus is almost the last plugin in the chain.
This results in histogram metrics being off as it will only start the timer when it reaches prometheus.

Make it the first plugin by default to make sure the metrics are correct.

erbesharat:
Adds caddy-prometheus middleware as the first middleware

fixes #169

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,10 @@ docker-run: image-build
 	docker run --name $(CONTAINER) $(IMAGE)
 
 docker-test:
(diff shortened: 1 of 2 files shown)


stp-ip:
I think this needs to happen within the run.go so that the middleware is called earlier from Caddy's side.

erbesharat:
Build fails because of the latest changes on the Caddy.

stp-ip:
Let's move to Go1.12 in Travis and our makefile to workaround the TLS1.3 issue with latest Caddy.

PR #226, opened 2019-02-27
[header]: Replace "Server" header value with "TXTDirect"

### LINKED ISSUE #221: [header] Replace "server" header with "TXTDirect"
feature

**What happened**:
When using caddy a server header is added "server: caddy".

Rewrite header to "Server: TXTDirect".

erbesharat:
Replaces "Server" header value with "TXTDirect"

fixes #221

### DIFF
--- a/caddy/caddy.go
+++ b/caddy/caddy.go
@@ -192,6 +192,8 @@ type Redirect struct {
 }
 
 func (rd Redirect) ServeHTTP(w http.ResponseWriter, r *http.Request) (int, error) {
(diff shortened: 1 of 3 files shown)


stp-ip:
As you mentioned we should increase our test coverage.
I think 3 tests would be helpful here:
* only standard headers + our server header present on requests
* using "header" plugin and test, if we can remove our Server header
* using "proxy" plugin and test, if we can remove our Server header

What do you think?

erbesharat:
Yeah, I think having testing using the caddy plugins is a good idea.

PR #251, opened 2019-03-20
[fallback] Add more tests and "www" support

### LINKED ISSUE #230: [fallback] Add additional tests
feature

**What happened**:
We are still running into fallback regressions.

We need more tests.

Test fallback, when only "www" or "host" enabled.

Currently enabling only "www" or "host" etc. results in a 500 error instead of the default redirect set globally.

erbesharat:
Adds more tests and "www" support

fixes #230

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -86,19 +86,19 @@ func fallback(w http.ResponseWriter, r *http.Request, fallback, recordType, fall
 		}
 	} else if contains(c.Enable, "www") {
 		s := strings.Join([]string{defaultProtocol, "://", defaultSub, ".", r.URL.Host}, "")
+		http.Redirect(w, r, s, code)
 		if c.Prometheus.Enable {
+			FallbacksCount.WithLabelValues(r.Host, recordType, "subdomain").Add(1)
 			RequestsByStatus.WithLabelValues(r.URL.Host, strconv.Itoa(code)).Add(1)
 		}
-		http.Redirect(w, r, s, code)
 	} else if c.Redirect != "" {
-		w.Header().Set("Content-Type", "")
-		w.Header().Set("Status-Code", strconv.Itoa(http.StatusForbidden))
+		w.Header().Set("Status-Code", strconv.Itoa(http.StatusMovedPermanently))
 
-		http.Redirect(w, r, c.Redirect, http.StatusForbidden)
+		http.Redirect(w, r, c.Redirect, http.StatusMovedPermanently)
 
 		if c.Prometheus.Enable {
-			FallbacksCount.WithLabelValues(r.Host, recordType, "global").Add(1)
-			RequestsByStatus.WithLabelValues(r.URL.Host, string(http.StatusForbidden)).Add(1)
+			FallbacksCount.WithLabelValues(r.Host, recordType, "redirect").Add(1)
+			RequestsByStatus.WithLabelValues(r.URL.Host, string(http.StatusMovedPermanently)).Add(1)
 		}
 	} else {
 		http.NotFound(w, r)
@@ -193,10 +193,8 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 
 	rec, err := getRecord(host, r.Context(), c, r)
 	if err != nil {
-		if err != nil {
-			fallback(w, r, "", "", "global", http.StatusMovedPermanently, c)
-			return nil
-		}
+		fallback(w, r, "", "", "global", http.StatusFound, c)
+		return nil
 	}
 
 	if !contains(c.Enable, rec.Type) {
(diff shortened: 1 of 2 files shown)


txtdirect.go
```diff
@@ -73,31 +73,37 @@ func contains(array []string, word string) bool {
...
+	if fallback != "" && fallbackType != "global" {
 		http.Redirect(w, r, fallback, code)
 		if c.Prometheus.Enable {
 			FallbacksCount.WithLabelValues(r.Host, recordType, fallbackType).Add(1)
 			RequestsByStatus.WithLabelValues(r.URL.Host, strconv.Itoa(code)).Add(1)
 		}
+	} else if contains(c.Enable, "www") {
+		s := strings.Join([]string{defaultProtocol, "://", defaultSub, ".", r.URL.Host}, "")
+		if c.Prometheus.Enable {
+			RequestsByStatus.WithLabelValues(r.URL.Host, strconv.Itoa(code)).Add(1)
```
stp-ip: Why no specific FallbacksCount metric?

txtdirect.go
```diff
@@ -73,31 +73,37 @@ func contains(array []string, word string) bool {
...
+		}
+		http.Redirect(w, r, s, code)
```
stp-ip: Let's redirect first and then add the metric.

txtdirect.go
```diff
@@ -73,31 +73,37 @@ func contains(array []string, word string) bool {
...
-			log.Printf("[txtdirect]: %s > %s", r.Host+r.URL.Path, c.Redirect)
-			w.Header().Set("Content-Type", "")
-			w.Header().Add("Status-Code", strconv.Itoa(http.StatusForbidden))
-			http.Redirect(w, r, c.Redirect, http.StatusForbidden)
-			if c.Prometheus.Enable {
-				FallbacksCount.WithLabelValues(r.Host, recordType, "global").Add(1)
-				RequestsByStatus.WithLabelValues(r.URL.Host, string(http.StatusForbidden)).Add(1)
-			}
+		w.Header().Set("Content-Type", "")
+		w.Header().Set("Status-Code", strconv.Itoa(http.StatusForbidden))
```
stp-ip: Why do we us StatusForbidden for the global fallback?
Also why add the specific content-type? Shouldn't this me a normal redirect using the global "redirect" config?

txtdirect.go
```diff
@@ -73,31 +73,37 @@ func contains(array []string, word string) bool {
...
+
+		http.Redirect(w, r, c.Redirect, http.StatusForbidden)
+
+		if c.Prometheus.Enable {
+			FallbacksCount.WithLabelValues(r.Host, recordType, "global").Add(1)
```
stp-ip: Using "global" ass fallbackType within the function, which includes both 404, "redirect" and "www" seems good, but I think we can split this up within the metrics:
404 -> no fallback count
redirect used -> "redirect"
www used -> "subdomain"

Or similar

txtdirect.go
```diff
@@ -181,51 +187,23 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
...
-					RequestsByStatus.WithLabelValues(host, strconv.Itoa(http.StatusMovedPermanently)).Add(1)
-				}
-				return nil
-			}
-			http.NotFound(w, r)
-			if c.Prometheus.Enable {
-				RequestsByStatus.WithLabelValues(host, strconv.Itoa(http.StatusNotFound)).Add(1)
-			}
+		if err != nil {
+			fallback(w, r, "", "", "global", http.StatusMovedPermanently, c)
```
stp-ip: Let's use temporary for the global fallback as this might be due to a misconfiguration so temporary seems best here.

txtdirect.go
```diff
@@ -181,51 +187,23 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
...
 	}
 
-	fallbackURL, code, err := getBaseTarget(rec, r)
-	if err != nil {
-		return err
-	}
+	fallbackURL, code := rec.To, rec.Code
 
 	if rec.Re != "" && rec.From != "" {
 		fallback(w, r, fallbackURL, rec.Type, "to", code, c)
```
stp-ip: We should log an error here that using both rec.Re and rec.From is not allowed.

txtdirect.go
erbesharat: Can't remember the reason we had for using StatusForbidden. It doesn't make sense to me too. What do you think it should be? 302 or the given code to the function? (Think the only place that we don't give a status code to fallback is after checking if r.host is an IP)

Also not sure about the Content-Type but we had this comment in our previous code:
`Empty Content-Type to prevent http.Redirect from writing an html response body`
But think it's actually opposite of the code that is inside net/http package.

https://github.com/golang/go/blob/master/src/net/http/server.go#L2090-L2096

stp-ip: Left over copy and paste code from the blacklist maybe?

stp-ip:
As far as I can see the recent failures are due to some networking issues with travis or something. Will merge and hope our google tests pass as they do locally.

PR #253, opened 2019-03-30
Inherit dependency versions from Caddy

stp-ip:
Remove manual version pinning from makefile as these are now set within Caddy itself.

**Release note**:

```release-note

```


stp-ip:
Good catch. Fixed and merged \o/

PR #262, opened 2019-04-23
[build]: Use Caddy's new build recommendations

erbesharat:
**What this PR does / why we need it**:
Use Caddy's new build recommendations

**Which issue this PR fixes**:
fixes #261

PR #265, opened 2019-04-24
Iterate modules

stp-ip:
* Simplifies caddy registration into setup.go files instead of separate caddy package, that needs importing
* Makes TXTDirect easier to import
* Adds cmd/txtdirect that is now easily go-gettable and also adds dependencies for our build, while using the main go module file

Builds on top of: #262 

Just an idea. Not sure, if that works perfectly. But that was the result after running into some issues with #262 

**Release note**:

```release-note

```


stp-ip:
Fixed the docs (non need for Mac specific docs anymore) and the license headers at the same time.

PR #268, opened 2019-05-06
[dockerv2]: (WIP) Fix image push

### LINKED ISSUE #214: [dockerv2] Pushing images doesn't seem to work
bug

**What happened**:
Pushing an image currently fails.

Pushing images using the vanity URL should work.

Needs a bit more investigation.

erbesharat:
Fixes dockerv2's image pushing

fixes #214

### DIFF
--- a/dockerv2.go
+++ b/dockerv2.go
@@ -49,9 +49,19 @@ func redirectDockerv2(w http.ResponseWriter, r *http.Request, rec record) error 
 		if err != nil {
 			return err
 		}
+
+		// Empty the RequestURI to prevent "http: Request.RequestURI can't be set in client requests" error
+		r.RequestURI = ""
+		r.URL = uri
+
+		resp, err := http.DefaultClient.Do(r)
+		if err != nil {
+			return err
+		}
+
 		w.Header().Add("Cache-Control", fmt.Sprintf("max-age=%d", status301CacheAge))
 		w.Header().Add("Status-Code", strconv.Itoa(http.StatusMovedPermanently))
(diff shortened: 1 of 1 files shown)


erbesharat:
Docker CLI stops working when it receives a redirect status code for blobs upload request.   
Creating and sending a new request instead of redirecting solved that problem but now it stops working when tries to send the `PUT MANIFEST (/v2/<name>/manifests/<reference>)` request and the Docker CLI returns `invalid checksum digest format` error.

It seems like it's because of the authentication. None of the requests contain an `Authorization` header. Don't know how we can handle the authentication process cause I couldn't find anything related in v2's docs.

PUT Manifest request description: https://docs.docker.com/registry/spec/api/#put-manifest
Image manifest upload step: https://docs.docker.com/registry/spec/api/#pushing-an-image-manifest

erbesharat:
I've tested the new code with `gcr.io` and it still returns an `invalid checksum digest format` error for pushing.

erbesharat:
Right now pulling an image with a specific tag works fine but pulling the latest image returns an error from Docker's API.

**Error Info:**

Request Path: `/v2/txtdirect/txtdirect/manifests/latest`
API Response:
```json
{
  "errors": [
    {
      "code": "MANIFEST_UNKNOWN",
      "message": "Failed to fetch \"latest\" from request \"/v2/txtdirect/txtdirect/manifests/latest\"."
    }
  ]
}
```
Response Content-Encoding: `gzip`

erbesharat:
I guess we can add a condition for gcr.io and trigger the fallback or just log something when the incoming path contains the latest tag.

PR #269, opened 2019-05-31
[qr]: Add support for `?qr`

### LINKED ISSUE #258: [idea] Create qr code for a short url
feature

**What happened**:
Currently for offline campaigns etc. one need to manually create qr codes.

Ability to easily get a qr code for a specific url with the help of a specific url or query key.

Just a rough idea and might not be necessary in the short term.
Various shorteners yourls and shlink both have this feature.
Usually the add `.qr` on the path or something similar.

erbesharat:
Adds support for `?qr`

fixes #258


setup_test.go
```diff
@@ -382,6 +382,51 @@ func TestCaddyParse(t *testing.T) {
 				},
 			},
 		},
+		{
+			`
+			txtdirect {
+				enable host qr
```
stp-ip: In my view enable should only be used to enable routing/type features not general features. Aka we don't enable prometheus, but we do enable host and path.

So enable qr shouldn't be needed.

qr.go
```diff
@@ -0,0 +1,87 @@
...
+	Qr, err := qrcode.New(r.Host+r.URL.String(), qrcode.High)
+	if err != nil {
+		return fmt.Errorf("Couldn't generate the Qr instance: %s", err.Error())
+	}
+	if err = qr.ParseColors(); err != nil {
+		return fmt.Errorf("Coudln't parse colors: %s", err.Error())
+	}
+	Qr.BackgroundColor, Qr.ForegroundColor = qr.backgroundColor, qr.foregroundColor
+
+	Qr.Write(qr.Size, w)
```
stp-ip: We should write a minimal html file instead. In Firefox for example a default image written back is transparent and therefore makes it hard to scan. Also helpful would be to add a clickable link to the actual webpage under the image so one can navigate there, if the qr code isn't necessary.

qr.go
```diff
@@ -0,0 +1,87 @@
...
+	fg, _ := hex.DecodeString(qr.ForegroundColor)
+	if err != nil {
+		return err
+	}
+	qr.foregroundColor = color.RGBA{fg[0], fg[1], fg[2], fg[3]}
+	return nil
+}
+
+func (qr *Qr) ParseQr(c *caddy.Controller) error {
+	switch c.Val() {
```
stp-ip: If we go with being able to configure these things (a lot of configuration options in the long run) we should expose the error recovery.

stp-ip:
Also needs a rebase as I was too slow in reviewing :(

qr.go
stp-ip: I can see the reason to not add html. It would make for easier embedding.

Let's discuss this before implementing a change.

erbesharat: Like the clickable link idea. But can't understand what you mean by easier embedding.

For the Firefox issue, we can add a custom white background color based on the request's `User Agent` header if writing HTML can't be considered.

stp-ip: Any reason not to default to having a custom white background color in general?

stp-ip: For context on the embedding.
Having the URL give back an image makes embedding into a different site etc. easier as it's just an image instead of playing with an iframe, when a full blown html would be given back. Does that make sense?

stp-ip:
So 2 things missing before merge:
* [ ] expose error recovery options for generated qr codes?
* [ ] custom white background?

Happy to discuss these.

erbesharat:
Seems like that we already have a default white background when the background color is empty in the config.
https://github.com/txtdirect/txtdirect/blob/8644cbd3b9abfc06604f299c34cb75d57be34ff2/qr.go#L75-L77

btw, I couldn't understand what you mean by exposing error recovery. can you please elaborate?

qr.go
erbesharat: I feel like dealing with iframe just makes the embedding harder. Think the image with a colored background should be enough and the user can add a link to the image while embedding it if they want it clickable.

stp-ip:
The library has a specific value for error recovery. It makes the QR code
have more data, but be easier to recover.

stp-ip:
Yeah let's go with image. The problem with the white background is that
it's transparent I think. At least when I tested in Firefox it was
transparent.

erbesharat:
The foreground was transparent, maybe that was the problem in Firefox. Changed the default foreground color and added recovery_level option to config.

PR #280, opened 2019-08-17
[header]: Add referer header

### LINKED ISSUE #273: Add referer header
feature

**What happened**:
Currently it's hard to track, where a user came in and which redirect they used for users.

Add referrer header to add this information more easily to analytics pages etc.

Add `referer` header such as: `TXTDirect: {hostonly}` or `{hostonly}`
Open for other suggestions.

erbesharat:
Adds `Referer` header and `ref=` field to record struct

fixes #273

### DIFF
--- a/txtdirect.go
+++ b/txtdirect.go
@@ -128,16 +128,6 @@ func isIP(host string) bool {
 func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 	w.Header().Set("Server", "TXTDirect")
 
-	// Add referer header
-	if r.Header.Get("Referer") == "" {
-		host := r.Host
-		if strings.Contains(host, ":") {
-			hostSlice := strings.Split(host, ":")
-			host = hostSlice[0]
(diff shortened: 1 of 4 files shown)


txtdirect.go
```diff
@@ -160,6 +160,16 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
 		return nil
 	}
 
+	// Add referer header
+	if rec.Ref && r.Header.Get("Referer") == "" {
```
stp-ip: In which case is the Referer header already set?

erbesharat: Maybe when the user sends the request from a search engine or something like that? Added this condition based on your comment on the issue.
https://github.com/txtdirect/txtdirect/issues/273#issuecomment-510471223

stp-ip: Ah yeah. Playing nice with other referers. Thanks for pointing that out. /lgtm.

PR #290, opened 2019-09-13
[record]: Implement multipart record parser

### LINKED ISSUE #288: Enable records longer than 255 chars
feature

**What happened**:
Currently we error on records being longer than 255 chars as this is the maximum safe length of a TXT record.

Allow multiple TXT records to be merged for TXTDirect to function accordingly.

Example:
```
"v=txtv0;type=host;re=^\\/([^\\/]*)\\/\\?$;https://cs.k8s.io/?q=$1&i=fosho&files=.*(%5E%5B%5Ev%5D.*OWNERS%24)%7C(%5Ev%5B%5Ee%5D.*OWNERS%24)%20%7C(%5Eve%5B%5En%5D.*OWNERS%24)%7C(%5Even%5B%5Ed%5D.*OWNERS%24)%7C(%5Evend%5B%5Eo%5D.*OWNERS%24)%7C(%5Evendo%5B%5Er%5D.*OWNERS%24)%7C(%5E%5B%5Ev%5D%5B%5Ee%5D%5B%5En%5D%5B%5Ed%5D%5B%5Eo%5D%5B%5Er%5D%2F.*OWNERS%24)%7C%5EOWNERS%24%7C%5EOWNERS_ALIASES%24&repos="
```
General idea would be:
* loop through all records and check for *v=txtvX* aka any version from TXTDirect
* sort them according to *p=X*
* default single records and records without *p=* to *p=1*
* merge each record key according to *p=X*

Idea:
```
"v=txtv0;type=host;re=^\\/([^\\/]*)\\/\\?$;to=https://cs.k8s.io/?q=$1&i=fosho&files=.*(%5E%5B%5Ev%5D.*OWNERS%24)%7C(%5Ev%5B%5Ee%5D.*OWNERS%24)%20%7C(%5Eve%5B%5En%5D.*OWNERS%24)%7C(%5Even%5B%5Ed%5D.*OWNERS%24)%7C(%5Evend%5B%5Eo%5D.*OWNERS%24)%7C(%5Evendo%5B%5Er%5D.*OWNERS%24)"
"p=2;to=%7C(%5E%5B%5Ev%5D%5B%5Ee%5D%5B%5En%5D%5B%5Ed%5D%5B%5Eo%5D%5B%5Er%5D%2F.*OWNERS%24)%7C%5EOWNERS%24%7C%5EOWNERS_ALIASES%24&repos="
```

erbesharat:
Implements multipart record parser

fixes #288 

Would love to hear thought on other approaches for the multipart parser. I don't like the current implementation because I feel it's too complicated to read and understand.

Also for merging the records in current implementation we have to use reflections and I personally don't like using it.

Remaining parts:

- [x] Merge records
- [x] Order records based on `p=`

**Release note**:

```release-note

```

### DIFF
--- a/record.go
+++ b/record.go
@@ -90,7 +90,7 @@ func (r *record) Parse(str string, w http.ResponseWriter, req *http.Request, c C
 		}
 
 		// Sort records based on their `p=` field
-		sort.Slice(&records, func(i, j int) bool {
+		sort.Slice(records, func(i, j int) bool {
 			return records[i].p < records[j].p
 		})
 
@@ -226,19 +226,23 @@ func parseMultipart(str string, w http.ResponseWriter, req *http.Request, c Conf
 	records = append(records, r)
 
 	// Look for the other parts
-	for i := 1; i <= strings.Count(str, "p="); i++ {
+	parts := strings.Count(str, "p=")
+	for i := 1; i <= parts; i++ {
 		// Find the current and next parts index
 		currentPart := strings.Index(str, "p=")
 		nextPart := strings.Index(str[currentPart+4:], "p=")
 
 		str = str[currentPart:]
-
-		if currentPart != nextPart && nextPart != -1 {
-			str = str[currentPart:nextPart]
-		}
-
 		s := strings.Split(str, ";")
 
+		// If the record has more than 2 parts, split the part between
+		if currentPart != nextPart && nextPart != -1 {
+			nextPart = strings.Index(str[4:], "p=")
+			s = strings.Split(str[:nextPart+4], ";")
+			str = str[nextPart+4:]
+		}
+
+		// Parse the record fields
 		var r record
 		for _, l := range s {
 			if err := r.parseFields(w, req, c, l); err != nil {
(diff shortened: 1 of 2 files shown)


erbesharat:
For more improvement, we can add a few type checking to reflection part to be sure nothing goes wrong. Think the buggiest part would be inside the `for` loop but not sure how we can add more checks and error handling to it. We can check if the `str` has a `p=` field inside it before every string indexing but that would probably make some performance issues.

stp-ip:
Is `p=` really needed? In the issue it's mentioned that most implementations should give back the TXT record ordered so that we could just merge them and splitting/merging is much much easier for the user.

erbesharat:
@stp-ip If we don't use `p=`, we have to either define a separator or force the user to end the record with a `;`. That's because the given string is already merged by the DNS server. For example, it would be simple to parse `to=https://example;to=.com` but `to=https://exampleto=.com` is a bit harder.

stp-ip:
For example, "to=https://example" ".com" would become "to=
https://example.com" as far as I can see from what you describe.

stp-ip:
If it is merged on the DNS server side already shouldn't it just work?

The user has to split it into multiple strings, but as long as it's merged
in the correct order we should be able to parse it and be done or?

erbesharat:
Yeah, it already works if it should be like the example in your comment. But I thought it should look like the example in the issue (field name specified in each record string). If the comment example is ok we can close the PR, keep the current code and only update the docs.

stp-ip:
Then I think it would be ok to update the docs and then keep the current code for now.

Sorry. Should have made it clear, that the issue description was the first thought and then the comment added the idea, that it might work simpler with just string splitting etc.

Will improve my issue descriptions.

PR #296, opened 2019-10-14
[path]: Add predefined regex mode

### LINKED ISSUE #256: [path] Add predefined regex mode
feature

**What happened**:
Currently a regex is applied using the path record. That works great for url shortener redirects.
For more complex redirect ideas using a list of regexes this might make the regex in path complicated.

One option is to add a new mode to path that has a predefined flow to lookup records to retrieve and match regexes.

`Path` would catch all paths and we would look up some predefined records.

_redirect.example.com <- path `v=txtv0;to=https://www.example.com/docs/;type=path`
_redirect.1.example.com <- first regex `v=txtv0;re=^/v[0-9]+\.[0-9]+(/.*)?$ to=https://www.example.com/docs/{1};type=host`
_redirect.2.example.com <- second regex `v=txtv0;re=^/(.*)$ to=https://www.example.com/docs/{1};type=host`

Does that make sense? I dislike adding another mode, but that would make it similar to how caddy and nginx do matching of specific redirects. Aka most specific wins.
We would basically iterate through subzones and then the most specific regex match would win and we would use it's specific rec.To.

So we would have our simply `from` for shorteners. More complex splitting using `re` and placeholders and a fully featured regex/redirect engine using the above advanced mode. Not sure how that would be configered in `path` yet.

This would need changes to `path` and `host`.

Would love your thoughts.

erbesharat:
Adds predefined regex mode

fixes #256 

### DIFF
--- a/fallback.go
+++ b/fallback.go
@@ -64,7 +64,7 @@ func fallback(w http.ResponseWriter, r *http.Request, fallbackType string, code 
 func (f *Fallback) countFallback(recType string) {
 	if f.config.Prometheus.Enable {
 		FallbacksCount.WithLabelValues(f.request.Host, recType, f.fallbackType).Add(1)
-		RequestsByStatus.WithLabelValues(f.request.URL.Host, string(f.code)).Add(1)
+		RequestsByStatus.WithLabelValues(f.request.URL.Host, strconv.Itoa(f.code)).Add(1)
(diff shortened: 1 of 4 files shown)


erbesharat:
Will add a few E2E tests then it would be ready for review.

PR #304, opened 2019-12-04
[validator]: Add the txtdirect-validator tool

### LINKED ISSUE #263: [spec] Add spec validation/monitoring tool.
feature

**What happened**:
Currently it's hard to monitor urls/redirects working correctly according to our spec.

Make it possible for a tool to read in the DNS and validate the correct upstream address and conformance to our spec.

Not sure on how to not reimplement txtdirect code for the monitoring tool so.

Maybe have a spec tool that reads in DNS text records and can validate, if and what version of our "spec" they conform to.
Better way to write a spec?
Monitoring tools usually can check the correct redirect so monitoring the redirect might not be necessary. Also it would be nice for a tool to read in a complete dns zone and then test conformance and monitor all URLs that are added there. That way we can confirm all urls still work as expected after a deploy or so.

Needs a lot more thoughts, but wanted to add the idea already.

erbesharat:
Exposes the record parser and adds the validator tool

fixes #263 

The features that might be good to have:
* A flag to pass a URI and use it for parsing the placeholders inside the record
* A flag to pass a config file and use it to parse a complete config (Right now it only lets the user customize enabled types).

**Release note**:
`nil`

### DIFF
--- a/cmd/txtdirect-validator/main.go
+++ b/cmd/txtdirect-validator/main.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+
+	"go.txtdirect.org/txtdirect"
+)
+
+func main() {
+	var types string
+	var enabled []string
+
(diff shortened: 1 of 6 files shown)


erbesharat:
Not sure what's the best way to write tests for the tool. We can call `go run` using exec and run the tool for each tests.

PR #309, opened 2019-12-19
[e2e]: Add tests for dockerv2

erbesharat:
Adds E2E tests for `dockerv2` type

Not ready yet, we have to add the custom registry as an insecure registry to the docker daemon config.

**Release note**:

```release-note

```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ cache:
 
 before_script:
   - docker pull k8s.gcr.io/coredns:1.6.2
-  - docker run -d -p 5000:5000 --name registry registry:2
+  - docker pull registry:2
 
 script:
   - go vet
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ cache:
 
 before_script:
   - docker pull k8s.gcr.io/coredns:1.6.2
+  - docker run -d -p 5000:5000 --name registry registry:2
 
 script:
   - go vet
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ cache:
 before_script:
   - docker pull k8s.gcr.io/coredns:1.6.2
   - docker pull registry:2
+  - go get -u github.com/google/go-containerregistry/cmd/crane
 
 script:
   - go vet
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,9 @@ docker-build:
 
 endtoend-test: docker-build
 	docker build -t $(IMAGE)-dirty .
-	[ ! "$(shell docker ps -a | grep registry)" ] && docker run -d -p 5000:5000 --name registry registry:2
 	cd e2e && \
 	docker build -t c.txtdirect.org/tester:dirty . && \
+ 	docker run -d -p 5000:5000 --name registry registry:2 && \
 	VERSION=$(VERSION)-dirty go run main.go
 
 version:
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ docker-build:
 
 endtoend-test: docker-build
 	docker build -t $(IMAGE)-dirty .
+	[ ! "$(shell docker ps -a | grep registry)" ] && docker run -d -p 5000:5000 --name registry registry:2
 	cd e2e && \
 	docker build -t c.txtdirect.org/tester:dirty . && \
 	VERSION=$(VERSION)-dirty go run main.go
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ endtoend-test: docker-build
 	docker build -t $(IMAGE)-dirty .
 	cd e2e && \
 	docker build -t c.txtdirect.org/tester:dirty . && \
- 	docker run -d -p 5000:5000 --name registry registry:2 && \
+ 	docker run -d -p 5000:5000 --name registry registry:2.7.1 && \
 	VERSION=$(VERSION)-dirty go run main.go
 
 version:
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -4,6 +4,6 @@ RUN apk --no-cache add git
 
 WORKDIR /e2e
 
-RUN GO111MODULE=on go get -u github.com/google/go-containerregistry/cmd/crane
+RUN GO111MODULE=on go get -u github.com/google/go-containerregistry/cmd/crane@34fb8ff
 
 CMD [ "go" ]
--- a/e2e/Dockerfile
+++ b/e2e/Dockerfile
@@ -4,4 +4,6 @@ RUN apk --no-cache add git
 
 WORKDIR /e2e
 
+RUN GO111MODULE=on go get -u github.com/google/go-containerregistry/cmd/crane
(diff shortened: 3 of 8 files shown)


stp-ip:
When all pulls and pushes go through crane, why does the docker daemon need to know about the unsecure registry?

erbesharat:
My bad, forgot to use crane instead of docker for pushing the test image to the registry.

.travis.yml
```diff
@@ -12,6 +12,8 @@ cache:
 
 before_script:
   - docker pull k8s.gcr.io/coredns:1.6.2
+  - docker pull registry:2
+  - go get -u github.com/google/go-containerregistry/cmd/crane
```
stp-ip: Why the need to pull in crane here? Isn't this pulled into the tester docker file?

erbesharat: Since we use crane to push a testing image to the custom registry in manager-script before running the tests, we need to get crane in here.

erbesharat: Sidenote: If we go with pushing the image to registry with docker instead of crane, we have to add the registry as an insecure registry in daemon config.

stp-ip: Fair point. I think we can use crane to push the image.
Happy with that approach of pre pushing an image. What is your view on self-contained tests so. Shouldn't the test init a push? To be fair we do prepare CoreDNS and the registry already.

Probably the usual way would be to mock more of the responses, but I like the real world tooling usage to run into real world issues at least in e2e tests. For unit tests mocking might be better.

erbesharat:
I'm having some trouble with writings tests for the push.
This is what I have in mind to do:
1. Pull the image that we pushed to the registry in manager-script before running the dockerv2 tests. (This should be done in dockerv2's testcase)
2. Tag the pulled image to the zone address so we can push it
3. Push the image using crane

But the issue is crane needs a tarball of the image to push it to the registry and we have to tag the image so we can push it using txtdirect. But to create a tarball for the image we have to use `docker save` command and I couldn't find anything in crane to do that.

So I guess we have two options:
1. Find a fix for dind issues so we can install docker in the tester image
2. Implement a subcommand in crane to create tarballs for images

erbesharat:
Tried to use the `crane export` command but still can't find out why it doesn't work with the custom registry. It doesn't return any errors and just stays frozen.

erbesharat:
The issue is crane only exports remote images into a tarball.
So when we pull the testing image and tag it to push it again to the registry, crane can't export the newly tagged image into a tarball.

stp-ip:
As we have the image in the registry wouldn't it be possible to use `crane copy` to test push?
To be fair this is not perfect as push will fail, if the pull itself fails. 

What's your stance on pushing the e2e docker push tests to a separate PR? That way we can merge this in to have test coverage for the currently supported dockerv2 parts. Will also look into why our current dockerv2 usage on c.txtdirect.org etc. isn't working as expected.

erbesharat:
I agree with a separate PR for the push tests. Have I missed anything or the current fallback tests and the 1 pull test are fine for now? Couldn't think of any other scenario to cover right now.

e2e/dockerv2/main.go
```diff
@@ -49,69 +54,104 @@ var tests = []test{
...
-		}
-
-		req.Header = test.headers
-
-		resp, err := client.Do(req)
-		if err != nil {
-			result[false] = append(result[false], test)
-			log.Printf("[%s]: Couldn't send the request: %s", test.name, err.Error())
-			continue
+		if test.fallback {
```
stp-ip: I like the restructure on the fallback.

e2e/Dockerfile
```diff
@@ -4,4 +4,6 @@ RUN apk --no-cache add git
 
 WORKDIR /e2e
 
+RUN GO111MODULE=on go get -u github.com/google/go-containerregistry/cmd/crane
```
stp-ip: This should be pinnned to a specific version I think.
Two options:
```
Create a go.mod you use to track your globally installed tools, such as in ~/global-tools/go.mod, and cd to that directory prior to running go get or go install for any globally installed tools.

Create a go.mod for each tool in separate directories, such as ~/tools/gorename/go.mod and ~/tools/goimports/go.mod, and cd to that appropriate directory prior to running go get or go install for the tool.
```

Makefile
```diff
@@ -36,6 +36,7 @@ endtoend-test: docker-build
 	docker build -t $(IMAGE)-dirty .
 	cd e2e && \
 	docker build -t c.txtdirect.org/tester:dirty . && \
+ 	docker run -d -p 5000:5000 --name registry registry:2 && \
```
stp-ip: Can we pin this to a specific docker registry version?

e2e/dockerv2/main.go
```diff
@@ -1,34 +1,38 @@
...
-	host string
-	path string
+	host  string
+	path  string
+	image string
 }
 
+type Kind string
+
+const (
```
stp-ip: Why the abstraction of Kind?

stp-ip:
Also still a draft PR

e2e/Dockerfile
erbesharat: Can't we control the versions using the @version at the end of `go get` command and manage them only inside the Dockerfile? Since we're only using crane's binary, I'm not sure if go.mod is needed.

e2e/dockerv2/main.go
erbesharat: Just thought it may look better than direct string comparison. Will change it if it's unnecessary.

Makefile
erbesharat: The currently available tags are:
- 2.7.1, 2.7, 2, latest
- 2.6.2, 2.6

Since "2.7.1, 2.7, 2" are all the same image, is it needed to switch the tag?

stp-ip: Question is how much we trust Docker not to break during minor versions.
We could be specific and use 2.7.1, but happy to keep it as is. Up to you.

e2e/Dockerfile
stp-ip: Good idea. Simpler and same effect. +1

e2e/dockerv2/main.go
stp-ip: It makes the overall code a bit harder to understand in my view as the string for each type wouldn't change frequently anyway. Ok to keep it as is, but feel like this might be an over optimization for nicer code > readability. What do you think?

Makefile
erbesharat: Yeah I agree that Docker may mess up something during minor versions.
Will change it to `2.7.1`.

e2e/dockerv2/main.go
erbesharat: Well, It would've been useful if we had lots of types and not only pull and push. And since they're not gonna get changed at all I guess going with strings instead of Kind type is better rn.

e2e/Dockerfile
erbesharat: Since crane doesn't have specific versions, I used `34fb8ff` which is the hash for insecure flag PR's merge commit.

PR #310, opened 2020-01-02
[caddy]: Overwrite Caddy's version

### LINKED ISSUE #267: Overwrite Caddy version with TXTDirect version in -version command
feature

**What happened**:
Currently we inherit the caddy version.

Use TXTDirect version in builds.

Already tried around a bit. Problem is caddy is using runtime information about the specific caddy module. Not sure we can really overwrite this.

erbesharat:
Reimplements the Caddy server runner and overwrites the version.

fixes #267 

Currently, the version is being read from `/txtdirectmain/version.go`. Should we change the version constant in this file to get the version from the `VERSION` file in the root of the repo?


stp-ip:
Using VERSION as the source feels better. We already have a list of various locations for version and adding another one that an be prevented seems good.

PR #311, opened 2020-01-04
[headers]: Add custom headers using records

### LINKED ISSUE #308: [header] Add header support to enable "HSTS Preload Support"
**Is this a BUG REPORT or FEATURE REQUEST?**: FEATURE REQUEST
bug | feature

**What happened**: I want to setup HSTS preload on a domain served by txtdirect.

**What you expected to happen**: A configuration option via DNS that lets me add HSTS Preload header.

1. Setup a redirect using txtd.io
2. Check the header
3. No HSTS header

```
content-type: text/html; charset=utf-8
date: Thu, 19 Dec 2019 16:38:43 GMT
location: 
server: txtd.io
status-code: 302
x-sid: txtd.io-eb58fd
```

**Anything else we need to know?**: An HSTS header is a fairly basic security requirement today to prevent ISPs from blocking your website, and a HSTS preload header is required to prevent the redirect from being hijacked.

Ref:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security
- https://hstspreload.org/

erbesharat:
Implements the feature to add custom headers using TXT records.

fixes #308 

Nan

**Release note**:

*Note: All the header values should be percent-encoded.*

Example for single record:
```
"_redirect.example.com.": "v=txtv0;to=https://example.com;type=host;>TestHeader=TestValue;code=302"
```

Example for chained records using path type:
```
"_redirect.path.example.com.": "v=txtv0;type=path;>TestHeader=TestValue;>TestHeader1=TestValue1"
"_redirect.child.path.example.com.": "v=txtv0;type=host;to=https://example.com;"
```

HSTS Example:
```
"_redirect.example.com.": "v=txtv0;to=https://example.com;type=host;>Strict-Transport-Security=max-age%3D63072000%3B+includeSubDomains%3B+preload+;code=302"
```

### DIFF
--- a/record.go
+++ b/record.go
@@ -34,6 +34,7 @@ type record struct {
 	Root    string
 	Re      string
 	Ref     bool
+	Headers map[string]string
 }
 
 // getRecord uses the given host to find a TXT record
@@ -71,12 +72,15 @@ func getRecord(host string, c Config, w http.ResponseWriter, r *http.Request) (r
 	return rec, nil
 }
 
-// Parse takes a string containing the DNS TXT record and returns
+// ParseRecord takes a string containing the DNS TXT record and returns
 // a TXTDirect record struct instance.
 // It will return an error if the DNS TXT record is not standard or
 // if the record type is not enabled in the TXTDirect's config.
 func ParseRecord(str string, w http.ResponseWriter, req *http.Request, c Config) (record, error) {
-	r := record{}
+	r := record{
+		Headers: map[string]string{},
+	}
+
 	s := strings.Split(str, ";")
 	for _, l := range s {
 		switch {
@@ -142,6 +146,9 @@ func ParseRecord(str string, w http.ResponseWriter, req *http.Request, c Config)
 			l = strings.TrimPrefix(l, "website=")
 			l = ParseURI(l, w, req, c)
 			r.Website = l
+		case strings.HasPrefix(l, ">"):
+			header := strings.Split(l, "=")
+			r.Headers[header[0][1:]] = header[1]
 
 		default:
 			tuple := strings.Split(l, "=")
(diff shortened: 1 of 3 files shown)


record.go
```diff
@@ -68,15 +69,25 @@ func getRecord(host string, c Config, w http.ResponseWriter, r *http.Request) (r
 
 	r = rec.addToContext(r)
 
+	// Add the headers from record to the response
```
stp-ip: Would this overwrite any headers set by the Caddyfile config?
I think we should document this here, what it overwrites. What do you think?

record_test.go
```diff
@@ -156,6 +156,29 @@ func TestParseRecord(t *testing.T) {
 			},
 			status: 404,
 		},
+		{
```
stp-ip: http headers are case insensitive. We should test this insensitivity and make it so.

stp-ip: Ref: http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

record.go
erbesharat: It actually overwrites any header with the same key, so we should either use `Add()` method instead, or document that it overwrites the previous headers.

I guess documenting makes more sense because adding the value may cause trouble in some cases.
Or we can use the `>` for overwriting and something like `>+` for adding another value to the header.

record_test.go
erbesharat: Aww completely missed that. I guess we have to loop through all the headers and lowercase their key and check them with the header provided in the record.

stp-ip: +1

record.go
stp-ip: Happy with overwrite for now + documentation.

The semantics of overwrite would work with ">" so adding addition later on would work.

PR #312, opened 2020-01-21
[WIP][general]: Restructure the project

erbesharat:
This PR would move almost every code we used before to their own subpackages and gives us the ability to easily use them outside the TXTDirect.

The QR and Prometheus plugins are currently directly attached to TXTDirect but I'm working on a mechanism like CoreDNS to automatically add plugins to the code so we can implement more external plugins in the future.

**Release note**:

```release-note

```


stp-ip:
Oh wow that's a big change. Will take some time to dive deep. Also probably pushing this after the next release aka after the 23rd.

erbesharat:
@stp-ip Definitely needs to be done in the next release. Just wanted to implement my ideas and keep them somewhere so I don't forget them :smile: Will add the [WIP] tag to the title since the plugin system isn't ready yet too.
Also, seems like the edge-case in predefined regexes is still there. Will fix that in a separate PR.

erbesharat:
I'm gonna close this PR since we've already started another refactor and the current codebase is tottally different from what we have here. Will refrence this PR later if anything from here got used during the current refactor.

PR #317, opened 2020-02-05
(path): Stop adding query to path by default

erbesharat:
If the custom regex doesn't need the request query and we add it by default, it will cause some problems and generate a wrong zone address. Fixed this issue by only adding the request query when a custom regex needs it.

**Release note**:

```release-note

```

### DIFF
--- a/path.go
+++ b/path.go
@@ -197,7 +197,12 @@ func zoneFromPath(r *http.Request, rec record) (string, int, []string, error) {
 		path = path[:strings.Index(path, "/info/refs")]
 	}
 
-	path = fmt.Sprintf("%s?%s", path, r.URL.RawQuery)
+	// Only add request query to path if the custom regex needs it. Unless it
+	// might cause problems on custom regexes that don't need query to generate
+	// the zone
+	if strings.Contains(rec.Re, "query") {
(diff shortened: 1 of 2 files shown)


path.go
```diff
@@ -197,7 +197,12 @@ func zoneFromPath(r *http.Request, rec record) (string, int, []string, error) {
 		path = path[:strings.Index(path, "/info/refs")]
 	}
 
-	path = fmt.Sprintf("%s?%s", path, r.URL.RawQuery)
+	// Only add request query to path if the custom regex needs it. Unless it
+	// might cause problems on custom regexes that don't need query to generate
+	// the zone
+	if strings.Contains(rec.Re, "query") {
```
stp-ip: Couldn't we add it to the context instead? That way we have it accessible, but not pollute path not even in the case when it's needed by the regex.

erbesharat: Not sure how adding it to the context can be helpful because either way, we have to merge path and query before passing it to the custom regex.

erbesharat: Sidenote: I think it would be good to work on a refactor for the `zoneFromPath` function in one of the next milestones because that part of the code is really hard to maintain right now.
We can move some parts of it to separate methods and add more unit tests for them to make sure everything is stable and works as expected and make the futures changes easier to implement.

PR #318, opened 2020-02-13
[record]: Add `use=` field

erbesharat:
Adds the `use=` field to record to provide a feature like how CNAME records work. The `use=` field can be used to redirect a TXT record request to another upstream zone. Basically use a TXT record that has been specified for another zone address.

~~Still trying to figure out how to fix the bugs on the `path` type, so not ready for a full review.~~

**Release note**:

```release-note

```

### DIFF
--- a/e2e/host/zonefile
+++ b/e2e/host/zonefile
@@ -10,24 +10,29 @@
 
 $TTL 1h
 
-example.com.                IN A 172.20.10.2
+example.com.                                     IN A 172.20.10.2
 
 ; Host records
 
-to.host.host.example.com.         IN A 172.20.10.2
-_redirect.to.host.host.example.com.          IN TXT "v=txtv0;to=https://to-redirect.host.host.example.com;type=host;code=302" 
+to.host.host.example.com.                        IN A 172.20.10.2
+_redirect.to.host.host.example.com.              IN TXT "v=txtv0;to=https://to-redirect.host.host.example.com;type=host;code=302" 
 
-nocode.host.host.example.com.    IN A 172.20.10.2
-_redirect.nocode.host.host.example.com.      IN TXT "v=txtv0;to=https://nocode.host.host.example.com;type=host"
+nocode.host.host.example.com.                    IN A 172.20.10.2
+_redirect.nocode.host.host.example.com.          IN TXT "v=txtv0;to=https://nocode.host.host.example.com;type=host"
 
-noversion.host.host.example.com. IN A 172.20.10.2
-_redirect.noversion.host.host.example.com.   IN TXT "to=https://noversion.host.host.example.com;type=host"
+noversion.host.host.example.com.                 IN A 172.20.10.2
+_redirect.noversion.host.host.example.com.       IN TXT "to=https://noversion.host.host.example.com;type=host"
 
-noto.host.host.example.com.      IN A 172.20.10.2
-_redirect.noto.host.host.example.com.        IN TXT "v=txtv0;type=host"
+noto.host.host.example.com.                      IN A 172.20.10.2
+_redirect.noto.host.host.example.com.            IN TXT "v=txtv0;type=host"
 
-referer.host.host.example.com.   IN A 172.20.10.2
-_redirect.referer.host.host.example.com.     IN TXT "to=https://referer-redirect.host.host.example.com;type=host;ref=true"
+referer.host.host.example.com.                   IN A 172.20.10.2
+_redirect.referer.host.host.example.com.         IN TXT "to=https://referer-redirect.host.host.example.com;type=host;ref=true"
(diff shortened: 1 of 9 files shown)


erbesharat:
Found an edge-case on predefined regex records. Let's not merge this one till it gets fixed.

e2e/main.go
```diff
@@ -177,14 +177,15 @@ func (d *dockerManager) StartContainers() error {
 }
 
 func (d *dockerManager) StopContainers() error {
-	if strings.Contains(d.dir, "dockerv2") {
-		_, err := exec.Command("docker",
+	if strings.Contains(d.dir, "path") {
```
stp-ip: Why the specific check here?

txtdirect.go
```diff
@@ -173,12 +180,31 @@ func Redirect(w http.ResponseWriter, r *http.Request, c Config) error {
...
 
 	rec, err := getRecord(host, c, w, r)
-	r = rec.addToContext(r)
 	if err != nil {
 		fallback(w, r, "global", http.StatusFound, c)
 		return nil
 	}
 
+	// Add the upstream zone address from the use= fields to the request context
+	// and replace the source record with the upstream record
```
stp-ip: We don't replace the source record with the upstream record anymore or?

e2e/host/zonefile
```diff
@@ -10,24 +10,29 @@
...
+_redirect.referer.host.host.example.com.         IN TXT "to=https://referer-redirect.host.host.example.com;type=host;ref=true"
 
-fallback.host.host.example.com.              IN A 172.20.10.2
-_redirect.fallback.host.host.example.com.    IN TXT "v=txtv0;to=https://{label30};type=host;code=302" 
+fallback.host.host.example.com.                  IN A 172.20.10.2
+_redirect.fallback.host.host.example.com.        IN TXT "v=txtv0;to=https://{label30};type=host;code=302" 
+
+sourcerecord.host.host.example.com.              IN A 172.20.10.2
+upstreamrecord.host.host.example.com.            IN A 172.20.10.2
+_redirect.sourcerecord.host.host.example.com.    IN TXT "v=txtv0;to=https://wrongredirect.host.host.example.com;type=host;code=302;use=_redirect.upstreamrecord.host.host.example.com"
```
stp-ip: Interesting edge case that `to=` and `use=` are used at the same time. So `use=` takes precedent. Seems valid.

stp-ip: Generally I feel like we should separately test this.

e2e/path/main.go
```diff
@@ -74,6 +74,22 @@ var tests = []test{
...
+	{
+		name: "Redirect using the use= field and upstream record",
+		args: data{
+			host: "sourcerecord.path.path.example.com",
+			path: "/testing",
+		},
+		expected: "https://upstream.path.path.example.com",
+	},
+	{
+		name: "Redirect to upstream record using multiple use= fields",
```
stp-ip: Might be good to mention testing a non working upstream record and fallthrough to the next use= field. The description doesn't seem to explain the whole test well in my view.

e2e/path/main.go
```diff
@@ -74,6 +74,22 @@ var tests = []test{
...
+		args: data{
+			host: "srcrecord.path.path.example.com",
+			path: "/testing",
+		},
+		expected: "https://upstream.path.path.example.com",
+	},
```
stp-ip: Adding an additional test, that uses `use=`, but that does go into fallback. Aka there is no valid upstream working.
That seems to be a good additional failure case.

e2e/main.go
erbesharat: Oops. Leftover from debugging. Will remove it :)

txtdirect.go
erbesharat: Removed the second line. Leftover from the previous implementation.

e2e/main.go
erbesharat: Debugging Leftover. :sweat_smile:

record.go
```diff
@@ -193,17 +193,20 @@ func ParseRecord(str string, w http.ResponseWriter, req *http.Request, c Config)
 		r.Code = http.StatusFound
 	}
 
-	if r.Type == "" {
-		r.Type = "host"
-	}
+	// Only apply rules and default to records that doesn't point to a upstream record
```
stp-ip: Good catch

stp-ip: Probably could be refactored into a defaulting function, but not sure if that makes it simpler.

stp-ip:
Godocit failed. We should rename Godocit step from "build" to "Godocit" so it's easier to see in the PRs.

also not sure, why it failed so. Any ideas why the nil pointer dereference?

PR #319, opened 2020-02-19
[E2E]: Fix predefined regex bugs and add E2E tests

erbesharat:
Fixes a few bugs in predefined regex feature and record parser.
Also adds E2E tests based on [k8s.io's nginx config](https://github.com/kubernetes/k8s.io/blob/master/k8s.io/configmap-nginx.yaml)

Only added parts of the config that had new regexes so we now have E2E tests for all of the unique regexes used in the nginx config.

**Release note**:

```release-note

```

### DIFF
--- a/e2e/gometa/main.go
+++ b/e2e/gometa/main.go
@@ -17,6 +17,7 @@ type test struct {
 	name     string
 	args     data
 	expected string
+	comment  string
 }
 
 var tests = []test{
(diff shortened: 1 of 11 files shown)


erbesharat:
Will rebase this branch after #318 gets merged.

stp-ip:
The regex specific fix should have its own PR as well as the added e2e tests from k8s.io

merging them here makes rolling back, reviewing as well as reasoning about it harder I think. Thoughts?

erbesharat:
Yeah, I agree and didn't really like how this PR got messy with different fixes. Right now I've even created more fixes and commits on this branch that I haven't pushed them yet.
Will first create PRs for different fixes then when all of them got merged, will create a new PR for the k8s tests.

erbesharat:
The new branches for predefined regex fix and e2e tests are ready. Will create PRs for them after "use=" PR gets merged because both of them rely on those changes.
Will close this PR so we can continue the work on the separated PRs.

PR #326, opened 2020-03-25
[path]: Fix chaining's request loop bug and add more E2E tests

erbesharat:
Fixes chaining's request loop bug and add more E2E tests

**Release note**:

```release-note

```

### DIFF
--- a/e2e/host/zonefile
+++ b/e2e/host/zonefile
@@ -10,38 +10,38 @@
 
 $TTL 1h
 
-example.com.                                     IN A 172.20.10.2
+example.com.                                           IN A 172.20.10.2
 
 ; Host records
 
-to.host.host.example.com.                        IN A 172.20.10.2
-_redirect.to.host.host.example.com.              IN TXT "v=txtv0;to=https://to-redirect.host.host.example.com;type=host;code=302" 
+to.host.host.example.com.                              IN A 172.20.10.2
+_redirect.to.host.host.example.com.                    IN TXT "v=txtv0;to=https://to-redirect.host.host.example.com;type=host;code=302" 
 
-nocode.host.host.example.com.                    IN A 172.20.10.2
-_redirect.nocode.host.host.example.com.          IN TXT "v=txtv0;to=https://nocode.host.host.example.com;type=host"
+nocode.host.host.example.com.                          IN A 172.20.10.2
+_redirect.nocode.host.host.example.com.                IN TXT "v=txtv0;to=https://nocode.host.host.example.com;type=host"
 
-noversion.host.host.example.com.                 IN A 172.20.10.2
-_redirect.noversion.host.host.example.com.       IN TXT "to=https://noversion.host.host.example.com;type=host"
+noversion.host.host.example.com.                       IN A 172.20.10.2
+_redirect.noversion.host.host.example.com.             IN TXT "to=https://noversion.host.host.example.com;type=host"
 
-noto.host.host.example.com.                      IN A 172.20.10.2
-_redirect.noto.host.host.example.com.            IN TXT "v=txtv0;type=host"
+noto.host.host.example.com.                            IN A 172.20.10.2
+_redirect.noto.host.host.example.com.                  IN TXT "v=txtv0;type=host"
 
-referer.host.host.example.com.                   IN A 172.20.10.2
-_redirect.referer.host.host.example.com.         IN TXT "to=https://referer-redirect.host.host.example.com;type=host;ref=true"
+referer.host.host.example.com.                         IN A 172.20.10.2
+_redirect.referer.host.host.example.com.               IN TXT "to=https://referer-redirect.host.host.example.com;type=host;ref=true"
 
-fallback.host.host.example.com.                  IN A 172.20.10.2
-_redirect.fallback.host.host.example.com.        IN TXT "v=txtv0;to=https://{label30};type=host;code=302" 
+fallback.host.host.example.com.                        IN A 172.20.10.2
+_redirect.fallback.host.host.example.com.              IN TXT "v=txtv0;to=https://{label30};type=host;code=302" 
 
-sourcerecord.host.host.example.com.              IN A 172.20.10.2
-upstreamrecord.host.host.example.com.            IN A 172.20.10.2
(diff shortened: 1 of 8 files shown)


e2e/host/zonefile
```diff
@@ -10,35 +10,42 @@
...
+uprecord.host.host.example.com.                        IN A 172.20.10.2
+_redirect.srcwithto.host.host.example.com.             IN TXT "v=txtv0;to=https://wrongredirect.host.host.example.com;type=host;code=302;use=_redirect.upstreamrecord.host.host.example.com" 
+_redirect.uprecord.host.host.example.com.              IN TXT "v=txtv0;to=https://upstream.host.host.example.com;type=host;code=302" 
+
+chaining.path.path.example.com.                        IN A 172.20.10.2
+chained.host.path.example.com.                         IN A 172.20.10.2
+_redirect.chaining.path.path.example.com.              IN TXT "v=txtv0;type=path;to=https://fallback-to.path.path.example.com;code=302"
+_redirect.releases.chaining.path.path.example.com.     IN TXT "v=txtv0;type=path;to=http://chained.host.path.example.com"
+_redirect._.chained.host.path.example.com.             IN TXT "v=txtv0;type=host;to=https://chained-redirect.host.path.example.com;code=302"
+_redirect.chained.host.path.example.com.               IN TXT "v=txtv0;type=host;to=https://chained-redirect.host.path.example.com;code=302"
```
stp-ip: Why the double record here? We wouldn't know, which one got hit in the test. So not sure, why we need the wildcard as well as the direct record.

e2e/path/main.go
```diff
@@ -104,6 +104,30 @@ var tests = []test{
...
+	{
+		name: "Fallback when wrong path is given to chained path records",
+		args: data{
+			host: "pathchain.path.path.example.com",
+			path: "/wrong",
+		},
+		expected: "https://fallback-unknown-path.path.path.example.com",
+	},
+	{
+		name: "Use regex plaecholders in a host record chained to a path record",
```
stp-ip: *placeholders

e2e/path/main.go
```diff
@@ -104,6 +104,30 @@ var tests = []test{
 		fallback: true,
 		status:   404,
 	},
+	{
+		name: "Fallback when path not given to chained path records",
```
stp-ip: Can you elaborate a bit more what your thinking was on that test? Mostly me not being sure about the description.

e2e/path/zonefile
```diff
@@ -51,4 +51,14 @@ _redirect._.uprecord.path.path.example.com.            IN TXT "v=txtv0;to=https:
 
 
 fallbackupstream.path.path.example.com.                IN A 172.20.10.2
-_redirect.fallbackupstream.path.path.example.com.      IN TXT "v=txtv0;type=path;use=_redirect.wrongupstream.path.path.example.com"
\ No newline at end of file
+_redirect.fallbackupstream.path.path.example.com.      IN TXT "v=txtv0;type=path;use=_redirect.wrongupstream.path.path.example.com"
+
+pathchain.path.path.example.com.                       IN A 172.20.10.2
+_redirect.pathchain.path.path.example.com.             IN TXT "v=txtv0;type=path;to=https://fallback-to.path.path.example.com;code=302"
+_redirect._.pathchain.path.path.example.com.           IN TXT "v=txtv0;type=path;re=/fallback/;to=https://fallback-unknown-path.path.path.example.com"
```
stp-ip: Why is the `re=/fallback/` necessary? That wouldn't be hit or?

path.go
```diff
@@ -78,6 +80,18 @@ func (p *Path) Redirect() *record {
 	}
 
 	if rec.Type == "path" {
+		// If the current record and the last record added to the context
+		// are equal, use the current record's to= field to find the final record
```
stp-ip: add a comment about the previous loop, if sameness isn't checked.

e2e/host/zonefile
erbesharat: Leftovers from when I was trying to reproduce the bug. The wildcard should be removed and we need "_redirect.chained.host.path.example.com" because "_redirect.releases.chaining.path.path.example.com" is pointed to it.

e2e/path/zonefile
erbesharat: `re=/fallback/` is just a random regex so we can test fallback being triggered when the regex from the chained record doesn't support the request's path.

e2e/path/main.go
erbesharat: Should've chosen a better description, just a test to make sure requests to a chained path record don't go further when the path is empty and go into fallback instead. Not sure if it's useful, we already have the same test but it's not for a chained record.

e2e/path/zonefile
stp-ip: So it does go into the regex path and then fallback. That's an assumption the test is making right as it doesn't really know, if it goes into fallback or just a wildcard rec.To redirect or am I missing something?

e2e/path/main.go
stp-ip: Ahh so it tries to test it prioritising the rec.To over the further chained records. Sounds good. Just needs a better description.

e2e/path/zonefile
erbesharat: The test is just trying to see if two path records are chained correctly and if the chained record's regex works on the path or not. And in this case, the test sends a wrong path and expects the chained record to trigger fallback with "to" type. I guess we can consider it a "path > path > rec.To fallback" test.

stp-ip: +1

e2e/path/main.go
```diff
@@ -121,13 +121,21 @@ var tests = []test{
...
-		name: "Use regex plaecholders in a host record chained to a path record",
+		name: "Use regex placeholders in a host record chained to a path record",
 		args: data{
 			host: "chaining.path.path.example.com",
 			path: "/host/",
 		},
 		expected: "https://redirect-host.host.path.example.com",
 	},
+	{
+		name: "Use regex placeholders in fallback trigerred by chained record",
```
stp-ip:
```suggestion
		name: "Use regex placeholders in fallback triggered by chained record",
```

PR #327, opened 2020-04-03
[e2e]: Add new tests based on k8s' nginx config

erbesharat:
Add new E2E tests based on the server blocks from k8s.io's nginx config

Nginx config: https://github.com/kubernetes/k8s.io/blob/master/k8s.io/configmap-nginx.yaml

**Release note**:

```release-note

```


e2e/path/zonefile
```diff
@@ -66,3 +60,59 @@ _redirect.host-regex.host.path.example.com.            IN TXT "v=txtv0;type=host
 chaining-regex.path.path.example.com.                  IN A 172.20.10.2
 _redirect.chaining-regex.path.path.example.com.        IN TXT "v=txtv0;type=path;to=https://fallback-to.path.path.example.com;code=302"
 _redirect._.chaining-regex.path.path.example.com.      IN TXT "v=txtv0;type=path;re=/random/;to=https://fallback.path.path.example.com/{1}"
+
+srcrecord.path.path.example.com.                            IN A 172.20.10.2
```
stp-ip, line 64: Why the need for both srcrecord as well as sourcerecord etc.?
Might need better naming to declare the difference. From the current name it feels like it should be the same. One abbreviated the other one not.

e2e/path/zonefile
```diff
@@ -66,3 +60,59 @@ _redirect.host-regex.host.path.example.com.            IN TXT "v=txtv0;type=host
...
+_redirect.predefined-multinumbered.host.path.example.com.   IN TXT "v=txtv0;type=path;root=https://index.host.path.example.com;re=record"
+_redirect.1.predefined-multinumbered.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://first-record.host.path.example.com{1};re=^/(v[0-9]+\.[0-9]+\.[0-9]+(-(alpha|beta|rc)\.[0-9]+)?/.*)$"
+_redirect.2.predefined-multinumbered.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://second-record.host.path.example.com/ci{2}/{3};re=^/ci(-cross)?/?(.*)$"
+_redirect.3.predefined-multinumbered.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://third-record.host.path.example.com{1};re=^/(.*)$"
+
+predefined-version.host.path.example.com.                   IN A 172.20.10.2
+_redirect.predefined-version.host.path.example.com.         IN TXT "v=txtv0;type=path;root=https://index.host.path.example.com;re=record"
+_redirect.1.predefined-version.host.path.example.com.       IN TXT "v=txtv0;type=host;to=https://first-record.host.path.example.com{1};re=^/v[0-9]+\.[0-9]+(/.*)?$"
+_redirect.2.predefined-version.host.path.example.com.       IN TXT "v=txtv0;type=host;to=https://second-record.host.path.example.com/docs/{1};re=^/(.*)$"
```
stp-ip, line 93: Are we testing the second record as well?

e2e/path/zonefile
```diff
@@ -66,3 +60,59 @@ _redirect.host-regex.host.path.example.com.            IN TXT "v=txtv0;type=host
...
+
+predefined-versionword.host.path.example.com.               IN A 172.20.10.2
+_redirect.predefined-versionword.host.path.example.com.     IN TXT "v=txtv0;type=path;root=https://index.host.path.example.com;re=record"
+_redirect.1.predefined-versionword.host.path.example.com.   IN TXT "v=txtv0;type=host;to=https://first-record.host.path.example.com/release-{2}/examples{3};re=^/v([0-9]+\.[0-9]+)(/.*)?$"
+_redirect.2.predefined-versionword.host.path.example.com.   IN TXT "v=txtv0;type=host;to=https://second-record.host.path.example.com/{1};re=^/(.*)$"
```
stp-ip, line 98: Are we testing the second record as well?

e2e/path/zonefile
```diff
@@ -66,3 +60,59 @@ _redirect.host-regex.host.path.example.com.            IN TXT "v=txtv0;type=host
...
+
+predefined-simpletospecific.host.path.example.com.             IN A 172.20.10.2
+_redirect.predefined-simpletospecific.host.path.example.com.   IN TXT "v=txtv0;type=path;root=https://index.host.path.example.com;re=record"
+_redirect.1.predefined-simpletospecific.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://first-record.host.path.example.com/{2}/{3};re=^/([^/]*)/(.*)$"
+_redirect.2.predefined-simpletospecific.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://second-record.host.path.example.com/{2};re=^/([^/]*)/?$"
```
stp-ip, line 103: Are we testing the second record as well?

e2e/path/zonefile
```diff
@@ -66,3 +60,59 @@ _redirect.host-regex.host.path.example.com.            IN TXT "v=txtv0;type=host
...
+_redirect.4.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/4;re=^/good-first-issue$"
+_redirect.5.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/5;re=^/help-wanted$"
+_redirect.6.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/6;re=^/needs-ok-to-test$"
+_redirect.7.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/7;re=^/oncall$"
+_redirect.8.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/8;re=^/oncall-hotlist$"
+
+predefined-release.host.path.example.com.             IN A 172.20.10.2
+_redirect.predefined-release.host.path.example.com.   IN TXT "v=txtv0;type=path;root=https://index.host.path.example.com;re=record"
+_redirect.1.predefined-release.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/{2}{3};re=^/([^/]*)(/.*)?$"
```
stp-ip, line 118: Why use a numbered predefined regex here?

e2e/path/zonefile
```diff
@@ -66,3 +60,59 @@ _redirect.host-regex.host.path.example.com.            IN TXT "v=txtv0;type=host
...
+_redirect.1.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/1;re=^/api-review$"
+_redirect.2.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/2;re=^/bot-commands$"
+_redirect.3.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/3;re=^/github-labels$"
+_redirect.4.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/4;re=^/good-first-issue$"
+_redirect.5.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/5;re=^/help-wanted$"
+_redirect.6.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/6;re=^/needs-ok-to-test$"
+_redirect.7.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/7;re=^/oncall$"
+_redirect.8.predefined.host.path.example.com. IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/8;re=^/oncall-hotlist$"
```
stp-ip, line 114: Why use a numbered predefined regex here?

stp-ip:
The things open would be the insecure get and the sigs.k8s.io one.
Sigs is using negative look ahead, which we can't support thanks to Go not supporting it within their regex: https://github.com/kubernetes/k8s.io/blob/master/k8s.io/configmap-nginx.yaml#L62
Not sure yet on how to work around this.

The other one I'm looking into is the insecure one. If everything else can be pushed out of the nginx config, I'm already happy. So let's finish the above comments and discuss both the sigs and insecure one in separate issues after we merge.

stp-ip:
The insecure config seems like a special case. So either we add a special case caddy config or leave it in the nginx. We will see how much we are allowed to substitute anyway :)

stp-ip:
Also to be fair the insecure option just returns and prints a warning to use the secure one. So might not be something to support directly in txtdirect, but use caddy or keep using nginx for. Not the edge case I feel we should support (yet).

e2e/gometa/main.go
```diff
@@ -52,6 +53,42 @@ var tests = []test{
...
+		name: "Redirect to gometa record using the custom regex and single path",
+		args: data{
+			host: "custom-regex.gometa.path.example.com",
+			path: "/txtdirect?go-get=1",
+		},
+		expected: "https://package.gometa.path.example.com/txtdirect",
+		comment:  "for: k8s.io",
+	},
+	// {
+	// 	name: "Redirect to host record if request doesn't have ?go-get=1",
```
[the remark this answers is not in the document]
erbesharat: As you mentioned in your next comment, it's commented because we can't use negative look ahead. Should we remove it or keep it with a explanation comment about negative look ahead issue?

e2e/gometa/zonefile
```diff
@@ -26,3 +26,13 @@ _redirect.fallback.gometa.path.example.com.    IN TXT "v=txtv0;type=path;to=http
 fallback-empty-to.gometa.path.example.com.     IN A 172.20.10.2
 _redirect.empty-to.gometa.path.example.com.    IN TXT "v=txtv0;to=wrong:/url.test;type=gometa"
 
+custom-regex.gometa.path.example.com.          IN A   172.20.10.2
+_redirect.custom-regex.gometa.path.example.com.                            IN TXT "v=txtv0;type=path;root=https://no-path.gometa.path.example.com{uri};re=^/(?P<repo>[^/]*)(?P<subpath>/.*)?$"
+_redirect._._.custom-regex.gometa.path.example.com.                        IN TXT "v=txtv0;type=gometa;to=https://package.gometa.path.example.com/{repo}"
+
+
+; repo-and-package.gometa.path.example.com.                            IN A   172.20.10.2
```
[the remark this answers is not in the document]
erbesharat: They're `sigs.k8s.io` records.

e2e/path/zonefile
erbesharat: Oops, missed this one when fixed the rebase conflicts.

e2e/path/zonefile
erbesharat: We can't use wildcards because there's a difference between how we extract the matches from the regex result.
In here we're fetching the first item from the matches:
https://github.com/txtdirect/txtdirect/blob/dcb70326ffb529872f4c77bdce3dfd8747f539d0/path.go#L265
But in predefined regex we're fetching all of the matches:
https://github.com/txtdirect/txtdirect/blob/dcb70326ffb529872f4c77bdce3dfd8747f539d0/path.go#L159

The first code is from 2 years ago and I can't really remember the reasoning behind it.

e2e/gometa/main.go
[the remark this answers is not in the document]
stp-ip: Remove it and open a separate issue. We have git for keeping around historic code and issues for things to keep in mind and work on I think

e2e/path/zonefile
stp-ip: Mhh that feels like something we should figure out and fix or document. Happy for you to create an issue and assign this to me or you :)

e2e/path/zonefile
erbesharat: Predefined regex seemed like the closest thing to the `rewrite` config from Nginx to me. Not sure what's a better approach for it.

stp-ip: Felt like this could looks like a url shortener type, which could be handled with the path and host types. As the longer the list grows the more regex matches and dns requests the server needs to do. For subdomain matching that is not the case. Maybe I'm missing something.

e2e/path/zonefile
```diff
@@ -90,28 +90,28 @@ _redirect.3.predefined-multinumbered.host.path.example.com. IN TXT "v=txtv0;type
...
+_redirect.api-review.predefined.host.path.example.com.         IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/1"
+_redirect.bot-commands.predefined.host.path.example.com.       IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/2"
+_redirect.github-labels.predefined.host.path.example.com.      IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/3"
+_redirect.good-first-issue.predefined.host.path.example.com.   IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/4"
+_redirect.help-wanted.predefined.host.path.example.com.        IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/5"
+_redirect.needs-ok-to-test.predefined.host.path.example.com.   IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/6"
+_redirect.oncall.predefined.host.path.example.com.             IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/7"
+_redirect.oncall-hotlist.predefined.host.path.example.com.     IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/8"
```
stp-ip, line 114: It's not predefined anymore. So we should probably change the sub zone to something closer to reality.

PR #334, opened 2020-04-22
[record]: Add apex zone wildcard check

### LINKED ISSUE #286: Pull in records for apex from "_" zone.
feature

**What happened**:
When using the apex zone for redirects we add a huge issue in terms of DNS access for the apex zone.

Enable clean apex links such as example.com/txtdirect without the impact of cluttering the apex zone.

We currently already use "_" for wildcard records.
We could look for records in general under the "_" subzone as well.

That adds a few more DNS requests, but might be worth it.
For the first implementation we could limit the additional lookup for apex zones. So only do a lookup for *._.example.com, when example.com is used, but not when sub.example.com is used.

Thoughts?

erbesharat:
Add a check to fetch the record from the apex zone's wildcard if the apex zone returned NXDOMAIN error.

fixes #286 

**Release note**:

```release-note

```

### DIFF
--- a/e2e/host/zonefile
+++ b/e2e/host/zonefile
@@ -48,3 +48,11 @@ chained.host.path.example.com.                         IN A 172.20.10.2
 _redirect.chaining.path.path.example.com.              IN TXT "v=txtv0;type=path;to=https://fallback-to.path.path.example.com;code=302"
 _redirect.releases.chaining.path.path.example.com.     IN TXT "v=txtv0;type=path;to=http://chained.host.path.example.com"
 _redirect.chained.host.path.example.com.               IN TXT "v=txtv0;type=host;to=https://chained-redirect.host.path.example.com;code=302"
+
+apex.host.host.example.com.                            IN A 172.20.10.2
+_redirect._.apex.host.host.example.com.                IN TXT "v=txtv0;type=host;to=https://apex-to.host.host.example.com;code=302"
+
+apexuse.host.host.example.com.                         IN A 172.20.10.2
+userec.host.host.example.com.                          IN A 172.20.10.2
+_redirect._.apexuse.host.host.example.com.             IN TXT "v=txtv0;use=_redirect._.userec.host.host.example.com"
+_redirect._.userec.host.host.example.com.              IN TXT "v=txtv0;type=host;to=https://apexuse-to.host.host.example.com;code=302"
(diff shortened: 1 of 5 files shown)


e2e/host/zonefile
```diff
@@ -47,4 +47,12 @@ chaining.path.path.example.com.                        IN A 172.20.10.2
...
 _redirect.releases.chaining.path.path.example.com.     IN TXT "v=txtv0;type=path;to=http://chained.host.path.example.com"
-_redirect.chained.host.path.example.com.               IN TXT "v=txtv0;type=host;to=https://chained-redirect.host.path.example.com;code=302"
\ No newline at end of file
+_redirect.chained.host.path.example.com.               IN TXT "v=txtv0;type=host;to=https://chained-redirect.host.path.example.com;code=302"
+
+apex.host.host.example.com.                            IN A 172.20.10.2
+_redirect._.apex.host.host.example.com.                IN TXT "v=txtv0;type=host;to=https://apex-to.host.host.example.com;code=302"
+
+apexuse.host.host.example.com.                         IN A 172.20.10.2
+userec.host.host.example.com.                          IN A 172.20.10.2
```
stp-ip, line 56: Why the need to add an IP to userec? Shouldn't the TXT record suffice?

e2e/path/zonefile
```diff
@@ -116,3 +116,10 @@ _redirect.oncall-hotlist.path-subdomain.host.path.example.com.     IN TXT "v=txt
 predefined-release.host.path.example.com.                          IN A 172.20.10.2
 _redirect.predefined-release.host.path.example.com.                IN TXT "v=txtv0;type=path;root=https://index.host.path.example.com;re=record"
 _redirect.1.predefined-release.host.path.example.com.              IN TXT "v=txtv0;type=host;to=https://predefined-regex.host.path.example.com/{2}{3};re=^/([^/]*)(/.*)?$"
+
+
+apexuse.path.path.example.com.                                     IN A 172.20.10.2
+userec.path.path.example.com.                                      IN A 172.20.10.2
```
stp-ip, line 122: Why the need to add an IP to userec? Shouldn't the TXT record suffice?

e2e/host/zonefile
erbesharat: Leftovers :sweat_smile:

PR #345, opened 2020-05-02
[placeholders]: Add "$" prefix for named regex placeholders

### LINKED ISSUE #333: [placeholders]: Default placeholders shouldn't overwrite custom placeholders
bug

**What happened**:
Currently, the default placeholders would overwrite the named regex placeholders in case of a name collision.

Use a new prefix for custom placeholders like named regex placeholders to avoid overwriting them.

Using the same names as default placeholders for named regex placeholder. like "host" and etc.

What is the best prefix to use? suggestions:
- `{_my-named-regex}` "_"
- `{RE_my-named-regex}` "RE_"
- `{$my-named-regex}` "$"

**Environment**:
- TXTDirect version:  `master`

erbesharat:
Adds "$" prefix for named regex placeholders

fixes #333 

**Release note**:

```release-note

```


stp-ip:
Holding this for the next release as this is a breaking change. Will merge after the release.

PR #349, opened 2020-09-02
(docs): update caddy config example for caddy 2

turbomaze:

The current example caddy file in the testing docs does not work with Caddy 2. This PR updates the testing docs to include a compatible caddyfile that seems to work as expected. I'm not familiar with Caddy so I'm sure it can be improved! If anyone is aware of improvements please let me know

### DIFF
--- a/docs/testing/README.md
+++ b/docs/testing/README.md
@@ -65,16 +65,21 @@ limitations under the License.
 
 - Create a caddyfile named `caddy.test`
   ```
-  example.test:8080 {
-      tls off
-      txtdirect {
-          enable path host
-          resolver 127.0.0.1:5353
-          logfile stdout
-      }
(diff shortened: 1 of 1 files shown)


docs/testing/README.md
```diff
@@ -65,16 +65,21 @@ limitations under the License.
...
-          resolver 127.0.0.1:5353
-          logfile stdout
-      }
-      log / stdout "{remote} - [{when_iso}] \"{method} {uri} {proto}\" {status} {size} {latency}"
-      errors stdout
+  {
+      order txtdirect last
   }
+
+  :8080
```
erbesharat, line 72: I understand that this is a standard and general example for default address, but since we're using `example.test` in zonefile examples and other places, it would be awesome if this could get changed.

erbesharat (changes requested):
Thanks for updating the docs and example. Wish v2 had better backward-compatibility for caddyfiles :smile: Everything else LGTM :+1:

PR #353, opened 2024-05-24
golang: 1.14 -> 1.22.3, alpine: 3.10 -> 3.20, adds mock environment via docker-compose

hwshadow:

* pulls txtdirect forward with updates to golang and alpine versions
* pulls dependencies forward as needed
* add docker compose mock environment that allows easy exploration of the project
* add some simple instructions for: building and running, with and without docker

* not sure how to run your test

**Release note**:
```release-note
golang: 1.14 -> 1.22.3, alpine: 3.10 -> 3.20, adds mock environment via docker-compose
```


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.10
+FROM docker.io/library/alpine:3.20
```
hwshadow, line 1: might be able to remove `docker.io/library/`, I was having issues building in WSL2, but I think I fixed my root issue, if you could try without the prefix, that would be grand.

